PR #20, opened 2018-06-27
Fix a typo in the helm chart

pureelk:
"lables" -> "labels".
This typo affects the helm upgrade command.


pure-k8s-plugin/templates/pure-provisioner.yaml
```diff
@@ -3,7 +3,7 @@ kind: Deployment
 metadata:
   name: pure-provisioner
   namespace: {{ .Release.Namespace }}
-  lables:
+  labels:
```
sun7927: Good catch :)

One more step is required to follow https://github.com/purestorage/helm-charts#updating-charts to build the package.

PR #21, opened 2018-06-27
match the flexPath to kubelet's volume-plugin-dir

sun7927:


patrick-east:
We need to bump the tag version, right?

PR #22, opened 2018-06-29
fixed a label issue which fails the upgrade, and recreate pure-flex pods

sun7927:


pure-k8s-plugin/templates/_helpers.tpl
```diff
@@ -2,9 +2,7 @@
 */}}
 {{- define "pure_k8s_plugin.labels" -}}
 generator: helm
-heritage: {{ .Release.Service | quote }}
```
taherv: What about version in the chart ? It seems to be removed ....

pure-k8s-plugin/README.md
```diff
@@ -102,7 +102,7 @@ helm upgrade pure-storage-driver pure/pure-k8s-plugin -f <your_own_dir>/yourvalu
 It's not recommended to upgrade by setting the `image.tag` in the image section of values.yaml, use the version of
 the helm repository with the tag version required. This will ensure the supporting changes are present in the templates.
 ```
-helm upgrade pure-storage-driver pure/pure-k8s-plugin -f <your_own_dir>/yourvalues.yaml --version <target version>
```
taherv: So recreating/restarting pods for flexvol is non-disruptive for any ongoing mounts. But what about the provisioner ? Can this cause some PVC to be not provisioned properly if the pod is recreated while provisioner was executing ?

pure-k8s-plugin/templates/_helpers.tpl
sun7927: Yes, It is removed. The version in labels will fail the upgrade to a different version, because when doing upgrade, it will check the labels by using the new version.

The chart version info can be found by "helm list".

pure-k8s-plugin/README.md
sun7927: Good point. ongoing PVC provisioning may be impacted. Customer should complete all the ongoing operations before upgrade.

patrick-east:
Please make pull requests from a fork and not branches on this repo.

PR #41, opened 2018-08-07
helm chart upgrade test script

sun7927:
This upgrade script is to test the upgrade from the any GA version to the current developing one.

This script can be run locally with pre-installed minikube, such as:
   `PRE_CHART_VERSION=latest test/run-k8s-plugin-upgrade-test.sh`

Also created a jenkins jobs for testing the upgrade, but not commit the jenkins job codes into this public repo.

http://re75jenkins.dev.purestorage.com/view/helm%20chart/


taherv:
How does one specify the version to upgrade from? In an environment variable?

PR #46, opened 2018-09-18
security improvement for plugin helm chart installation

sun7927:
I have manually tested it successfully in both kubernetes and openshift cluster

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,23 +12,25 @@ helm repo update
 ## Helm Setup
 Please ensure you have a correct service account with cluster-admin role in K8s/Openshift for Helm. 
 
-For K8s, if you hit the following error, solve it as
+1. Install Helm client and tiller under a namespace(i.e. project in openshift)
 ```
-Error:
-# helm list
-Error: configmaps is forbidden: User "system:serviceaccount:kube-system:default" cannot list configmaps in the namespace "kube-system"
+export TILLER_NAMESPACE=tiller
+# If helm is not installed, you can install as:
+curl -s https://storage.googleapis.com/kubernetes-helm/helm-v2.9.1-linux-amd64.tar.gz | tar xz
+cd linux-amd6a
 
-Solve it by:
-# kubectl create clusterrolebinding add-on-cluster-admin --clusterrole=cluster-admin --serviceaccount=kube-system:default
-```
+# Check for helm version:
+./helm version
 
-For Openshift, you may need extra step to configure RBAC for plugin
-```
-Configure RBAC for helm/tiller for installing
-# oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:kube-system:default
+# Initialize the helm client and tiller, and configure the correct role for it
+# For K8s:
+./helm init
+kubectl create clusterrolebinding add-on-cluster-admin --clusterrole=cluster-admin --serviceaccount=${TILLER_NAMESPACE}:default
 
-Configure RBAC for plugin (assume it is installed in project of "yourproject")
-# oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:yourproject:default
+# For Openshift:
+./helm init --client-only
+oc process -f https://github.com/openshift/origin/raw/master/examples/helm/tiller-template.yaml -p TILLER_NAMESPACE="${TILLER_NAMESPACE}" -p HELM_VERSION=v2.9.1 | oc create -f -
+oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:${TILLER_NAMESPACE}:tiller
 ```
 
 For more details, please see https://docs.helm.sh/using_helm/
--- a/pure-k8s-plugin/README.md
+++ b/pure-k8s-plugin/README.md
@@ -28,12 +28,13 @@ The following table lists the configurable parameters and their default values.
 
 |             Parameter       |            Description             |                    Default                |
 |-----------------------------|------------------------------------|-------------------------------------------|
+| `project`                   | The project or namespace where the plugin will be installed | `pure-storage-plugin` |
 | `image.name`                | The image name       to pull from  | `purestorage/k8s`                         |
 | `image.tag`                 | The image tag to pull              | `2.1.0`                                   |
 | `image.pullPolicy`          | Image pull policy                  | `IfNotPresent`                            |
 | `app.debug`                 | Enable/disable debug mode for app  | `false`                                   |
 | `storageclass.isPureDefault`| Set `pure` storageclass to the default | `false`                               |
-| `clusterrolebinding.serviceAccount.name`| Name of K8s service account for app | `default`                    |
+| `clusterrolebinding.serviceAccount.name`| Name of K8s/openshift service account for installing the plugin | `installer`                    |
 | `flasharray.sanType`        | Block volume access protocol, either ISCSI or FC | `ISCSI`                     |
 | `flasharray.defaultFSType`  | Block volume default filesystem type. *Not recommended to change!* | `xfs`     |
 | `flasharray.defaultFSOpt`  | Block volume default mkfs options. *Not recommended to change!* | `-q`          |
@@ -71,6 +72,9 @@ arrays:
         rack: "6a"
 ```
 
+## Install the plugin in a separate namespace(i.e. project)
+For security reason, it's strongly recommended to install the plugin in a new separated namespace/project.
+
 Customize your values.yaml including arrays info (replacement for pure.json), and then install with your values.yaml. Better to set a release name such as "pure-storage-driver"
 
 Dry run the installation, and make sure your values.yaml working correctly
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -12,23 +12,25 @@ helm repo update
...
-```
-Configure RBAC for helm/tiller for installing
-# oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:kube-system:default
-
-Configure RBAC for plugin (assume it is installed in project of "yourproject")
-# oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:yourproject:default
+export TILLER_NAMESPACE=tiller
+# If helm is not installed, you can install as:
+curl -s https://storage.googleapis.com/kubernetes-helm/helm-v2.9.1-linux-amd64.tar.gz | tar xz
+cd linux-amd6a
```
taherv: amd6a ?

README.md
```diff
@@ -12,23 +12,25 @@ helm repo update
...
+
+# Check for helm version:
+./helm version
```
taherv: What is this check ? What is the audience supposed to check ?

README.md
```diff
@@ -12,23 +12,25 @@ helm repo update
...
+
+# Initialize the helm client and tiller, and configure the correct role for it
+# For K8s:
+./helm init
+kubectl create clusterrolebinding add-on-cluster-admin --clusterrole=cluster-admin --serviceaccount=${TILLER_NAMESPACE}:default
+
+# For Openshift:
+./helm init --client-only
+oc process -f https://github.com/openshift/origin/raw/master/examples/helm/tiller-template.yaml -p TILLER_NAMESPACE="${TILLER_NAMESPACE}" -p HELM_VERSION=v2.9.1 | oc create -f -
```
taherv: > -p HELM_VERSION=v2.9.1
Why is this required ?
And why is this required only for openshift ?

pure-k8s-plugin/README.md
```diff
@@ -28,12 +28,13 @@ The following table lists the configurable parameters and their default values.
 
 |             Parameter       |            Description             |                    Default                |
 |-----------------------------|------------------------------------|-------------------------------------------|
+| `project`                   | The project or namespace where the plugin will be installed | `pure-storage-plugin` |
```
taherv: Consider changing the default ... we might use this project for other pods/software.

pure-k8s-plugin/README.md
```diff
@@ -28,12 +28,13 @@ The following table lists the configurable parameters and their default values.
...
 | `image.name`                | The image name       to pull from  | `purestorage/k8s`                         |
 | `image.tag`                 | The image tag to pull              | `2.1.0`                                   |
 | `image.pullPolicy`          | Image pull policy                  | `IfNotPresent`                            |
 | `app.debug`                 | Enable/disable debug mode for app  | `false`                                   |
 | `storageclass.isPureDefault`| Set `pure` storageclass to the default | `false`                               |
-| `clusterrolebinding.serviceAccount.name`| Name of K8s service account for app | `default`                    |
+| `clusterrolebinding.serviceAccount.name`| Name of K8s/openshift service account for installing the plugin | `installer`                    |
```
taherv: What does the default value "installer" mean ?

pure-k8s-plugin/README.md
```diff
@@ -71,6 +72,9 @@ arrays:
         rack: "6a"
 ```
 
+## Install the plugin in a separate namespace(i.e. project)
+For security reason, it's strongly recommended to install the plugin in a new separated namespace/project.
```
taherv: separated -> separate ?

pure-k8s-plugin/README.md
```diff
@@ -105,7 +120,7 @@ helm upgrade pure-storage-driver pure/pure-k8s-plugin -f <your_own_dir>/yourvalu
 It's not recommended to upgrade by setting the `image.tag` in the image section of values.yaml, use the version of
 the helm repository with the tag version required. This will ensure the supporting changes are present in the templates.
 ```
-helm upgrade pure-storage-driver pure/pure-k8s-plugin -f <your_own_dir>/yourvalues.yaml --version <target version> --recreate-pods
+helm upgrade pure-storage-driver pure/pure-k8s-plugin -f <your_own_dir>/yourvalues.yaml --version <target version>
```
taherv: How/Where would I get "targetversion" from ....

pure-k8s-plugin/templates/pure-flex-daemon.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: extensions/v1beta1
 kind: DaemonSet
 metadata:
   name: pure-flex
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.project }}
```
taherv: What does "project" mean for the helm chart when it is used to install in  Kubernetes (not openshift) cluster ?

pure-k8s-plugin/templates/pure-flex-daemon.yaml
```diff
@@ -17,6 +17,11 @@ spec:
         app: pure-flex
 {{ include "pure_k8s_plugin.labels" . | indent 8}}
     spec:
+{{- if (eq "openshift" .Values.orchestrator.name) }}
```
taherv: What is this ? And why is it required ?

pure-k8s-plugin/values.yaml
```diff
@@ -2,6 +2,9 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# which openshift project or k8s namespace will the plugin be installed
+project: pure-storage-plugin
```
taherv: Shouldnt this be called "namespace" for kubernetes users who are not openshift ?

pure-k8s-plugin/templates/pure-flex-daemon.yaml
sun7927: it will deploy the flex plugin only on compute nodes. Without it, the openshift's namespace must be configured with the correct nodeSelector to deploy the flex plugin on the correct nodes.

pure-k8s-plugin/templates/pure-flex-daemon.yaml
sun7927: will change it to namespace which can be understood by both openshift and kubernetes.

pure-k8s-plugin/README.md
sun7927: how to find the <target version>:
   1. helm repo update
   2. helm search pure-k8s-plugin -l

will add into the README.md

pure-k8s-plugin/README.md
sun7927: "installer" is a serviceaccount will be created to install the plugin.

pure-k8s-plugin/README.md
sun7927: will change it to "pure-storage-driver"

README.md
sun7927: helm installation in openshift is a little different from the kubernetes.
https://blog.openshift.com/getting-started-helm-openshift/

README.md
sun7927: it's to get the helm version which is required to install the correct version of helm server(tiller).

README.md
sun7927: it's for Linux. I will add more info for MacOS and Windows

taherv:
I feel like we should not re-document helm install .... those instructions can get obsolete pretty quickly. Helm must have some "install" or "getting started" for kubernetes/openshift readme/blog. We should just reference that.
That will also serve as a check for us. If we have to do something different than what helm suggests, why are we different ? Is the difference really required ?

pure-k8s-plugin/templates/pure-flex-daemon.yaml
patrick-east: Would it make sense to expose the node selector through values.yaml so that a deployer could limit which nodes have the flex plugin installed on? We can default to reasonable things but giving someone the flexibility there seems useful.

pure-k8s-plugin/values.yaml
```diff
@@ -19,7 +19,7 @@ storageclass:
 # specify the service account name for this app
 clusterrolebinding:
   serviceAccount:
-    name: default
+    name: installer
```
patrick-east: Why installer? This service account isn't just for installation afaik. I would vote for something like `pure` as the default

pure-k8s-plugin/values.yaml
```diff
@@ -38,6 +38,8 @@ namespace:
 # if you want to install flex into a different place, you need to
 # overwrite the flexpath.
 orchestrator:
+  # which openshift project or k8s namespace will the plugin be installed
+  namespace: pure-storage-driver
```
patrick-east: Why are we defining the namespace here? Helm already has a setting to define namespaces `Release.Namespace` which is set via the install command, from the docs `--namespace string         namespace to install the release into. Defaults to the current kube config namespace.`

We can just key off of that in our templates where we set the namespace.

Either way we need to add documentation that explains where things get deployed.

patrick-east:
Can you update the PR description to include what problem this is solving and how it solves it? That information will be useful for others that are looking at changes later in the repo.

README.md
```diff
@@ -12,23 +12,19 @@ helm repo update
...
+2. For Openshift
+https://blog.openshift.com/getting-started-helm-openshift/
 
-For Openshift, you may need extra step to configure RBAC for plugin
-```
-Configure RBAC for helm/tiller for installing
-# oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:kube-system:default
+# After installation, configure the correct role for it
+# For K8s:
+kubectl create clusterrolebinding add-on-cluster-admin --clusterrole=cluster-admin --serviceaccount=${TILLER_NAMESPACE}:default
```
patrick-east: This probably needs more documentation around what it is actually doing. Answering the question of "Why am i giving this thing admin permissions?" will probably save time

pure-k8s-plugin/values.yaml
sdodsley: I agree the service account name should be something to identify it to be Pure related

pure-k8s-plugin/templates/pure-flex-daemon.yaml
sdodsley: doesn't the flex plugin only install on nodes with the compute flag set anyway? Seems to do that in OpenShift 3.9 and higher, so I'm assuming the same for k8s 1.9 or higher.

README.md
sdodsley: I agree we need an explanation anywhere we are changing any sort of security mechanisma.

pure-k8s-plugin/templates/pure-flex-daemon.yaml
sun7927: Yes, we can expose the node selector, however it has to work with the namespace's node selector together. It makes more sense to me not to expose it through the values.yaml. Instead, deployer should use namespace's node selector.

pure-k8s-plugin/values.yaml
sun7927: make sense, will change.

pure-k8s-plugin/values.yaml
sun7927: will replace the defined namespace with Release.Namespace.

PR #51, opened 2018-12-20
Add pure-rawblock storage class

dinathom:
This is needed for rawblock support.

### DIFF
--- a/pure-k8s-plugin/templates/storageclass.yaml
+++ b/pure-k8s-plugin/templates/storageclass.yaml
@@ -32,3 +32,15 @@ metadata:
 provisioner: pure-provisioner
 parameters:
     backend: block
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: pure-rawblock
+  labels:
+    kubernetes.io/cluster-service: "true"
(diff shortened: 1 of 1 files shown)


pure-k8s-plugin/templates/storageclass.yaml
```diff
@@ -32,3 +32,15 @@ metadata:
...
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: pure-rawblock
+  labels:
+    kubernetes.io/cluster-service: "true"
+{{ include "pure_k8s_plugin.labels" . | indent 4}}
+provisioner: pure-provisioner
+parameters:
+    backend: block
```
taherv: volumemode is a new parameter ?
Can we do backend: rawblock (and not have volumemode) ?

dinathom: Per our discussion backend=block is used by the provisioner and we need the second parameter to  choose our mount-options by pure-flex. So will continue using a 2nd parameter volumemode=raw to indicate rawblock. 
I prefer using volumemode=raw over mountmode=raw because of the relation to k8s volumemode

dinathom:
Can I push this PR into the repo so that Simon can start running his tests?

PR #53, opened 2019-01-17
Add option to set  storageclass backend type

japaniel:
In addition to setting `pure` storageclass as default storageclass we desired the ability to choose which type of backend the default was. These changes allow you to set the default backend type of `block` or `file`

An example would be to edit the `values.yaml` from:
```
# do you want to set pure as the default storageclass?
storageclass:
  isPureDefault: false
```
to: 
```
# do you want to set pure as the default storageclass?
storageclass:
  isPureDefault: false
  pureBackend: file
```

### DIFF
--- a/pure-k8s-plugin/templates/storageclass.yaml
+++ b/pure-k8s-plugin/templates/storageclass.yaml
@@ -9,7 +9,7 @@ metadata:
 {{ include "pure_k8s_plugin.labels" . | indent 4}}
 provisioner: pure-provisioner
 parameters:
-    backend: block
+    backend: {{ default "block" .Values.storageclass.pureBackend }}
 ---
 kind: StorageClass
 apiVersion: storage.k8s.io/v1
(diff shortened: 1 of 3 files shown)


patrick-east:
Thanks for the contribution!

Can you add this to the default values.yaml too? https://github.com/purestorage/helm-charts/blob/master/pure-k8s-plugin/values.yaml#L16

taherv:
@dinathom Please review.

pure-k8s-plugin/templates/storageclass.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
 {{ include "pure_k8s_plugin.labels" . | indent 4}}
 provisioner: pure-provisioner
 parameters:
-    backend: block
+    backend: {{ default "block" .Values.storageclass.pureBackend }}
```
dinathom: Since quoting strings is in the best practices might as well do: 
    backend: {{ .Values.storageclass.pureBackend | default "block" | quote }}

PR #59, opened 2019-01-23
Update orchestrator.openshift.flexPath parameter

sdodsley:
Point to the correct directory that has been used in OpenShift since 3.9. This now matches upstream kubernetes.

This is the major cause of customer issues - ie. not having this directory value correct, even though the README does say to check...

### DIFF
--- a/pure-k8s-plugin/values.yaml
+++ b/pure-k8s-plugin/values.yaml
@@ -46,7 +46,7 @@ orchestrator:
     # flexBaseDir is for image.tag < 2.0
     flexBaseDir: /etc/origin/node/kubelet-plugins
     # flexPath is for image.tag >= 2.0
-    flexPath: /etc/origin/node/kubelet-plugins/volume/exec
+    flexPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
   k8s:
     flexBaseDir: /usr/libexec/kubernetes/kubelet-plugins
     flexPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
--- a/pure-k8s-plugin/README.md
+++ b/pure-k8s-plugin/README.md
@@ -45,7 +45,7 @@ The following table lists the configurable parameters and their default values.
 | `orchestrator.k8s.flexBaseDir` | Base path of directory to install flex plugin, works with orchestrator.name=k8s and image.tag < 2.0. Sub-dir of "volume/exec/pure~flex" will be automatically created under it | `/usr/libexec/kubernetes/kubelet-plugins` |
 | `orchestrator.k8s.flexPath` | Full path of directory to install flex plugin, works with orchestrator.name=k8s and image.tag >= 2.0.1 | `/usr/libexec/kubernetes/kubelet-plugins/volume/exec` |
 | `orchestrator.openshift.flexBaseDir` | Base path of directory to install flex plugin, works with orchestrator.name=openshift and image.tag < 2.0. Sub-dir of "volume/exec/pure~flex" will be automatically created under it | `/etc/origin/node/kubelet-plugins` |
-| `orchestrator.openshift.flexPath` | Full path of directory to install flex plugin, works with orchestrator.name=k8s and image.tag >= 2.0.1 | `/etc/origin/node/kubelet-plugins/volume/exec` |
+| `orchestrator.openshift.flexPath` | Full path of directory to install flex plugin, works with orchestrator.name=k8s and image.tag >= 2.0.1 | `/usr/libexec/kubernetes/kubelet-plugins/volume/exec` |
 | *`arrays`                    | Array list of all the backend FlashArrays and FlashBlades | must be set by user, see an example below                |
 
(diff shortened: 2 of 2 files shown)


pure-k8s-plugin/values.yaml
```diff
@@ -46,7 +46,7 @@ orchestrator:
     # flexBaseDir is for image.tag < 2.0
     flexBaseDir: /etc/origin/node/kubelet-plugins
     # flexPath is for image.tag >= 2.0
-    flexPath: /etc/origin/node/kubelet-plugins/volume/exec
+    flexPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
```
taherv: Is there a way to use this path by checking the version of openshift ?
Something like 
```
if .openshift.version == 3.11+
     flexPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
else
     flexPath: /etc/origin/node/kubelet-plugins/volume/exec
```

taherv: Maybe change this to : 
```
flexPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
orchestrator : 
   name: openshift/k8s
.....
```

sdodsley: Getting the OpenShift version is pretty easy as we are installing in an existing environment, using the `oc version` command, but how that translates into the YAML file I don't know. The output of the command is something like:
```[root@sn1-pool-c07-04 ~]# oc version
oc v3.11.51
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://sn1-pool-c07-11.puretec.purestorage.com:8443
openshift v3.11.16
kubernetes v1.11.0+d4cacc0
```

for your second comment, I think that is probably deeper into the whole plugin architecture, so I would leave your engineering people to work that out.

taherv: I don't follow. The change I requested in the second comment was just to take a property out of an 'if' condition, because the value is the same in if (k8s) and if (openshift). If we make this change, we have the value in only one place, helps maintainability of the code.
The extra fixup needed then, is to replace orchestrator.flexPath with just flexPath in the chart templates ... and you are good to go. Small change now, but helps prevents bugs in the future.

taherv:
Apart from the nit I have, looks good to me.
Ok to merge, if tests pass.

dinathom:
The location depends on which RHEL os version is running.

See here: https://docs.openshift.com/container-platform/3.11/install_config/persistent_storage/persistent_storage_flex_volume.html

Quoting : 
In OpenShift Container Platform 3.11, since controller-manager runs as a static pod, the FlexVolume binary file that performs the attach and detach operations must be a self-contained executable file with no external dependencies.

On Atomic hosts, the default location of the FlexVolume plug-in directory is /etc/origin/kubelet-plugins/. You must place the FlexVolume executable file in the /etc/origin/kubelet-plugins/volume/exec/<vendor>~<driver>/<driver> directory on all master and nodes in the cluster.

pure-k8s-plugin/values.yaml
dinathom: @sdodsley @taherv from my reading this PR will not be a complete fix. It will break Atomic Host installations of OCP.

sun7927: We should have a better document to cover all the cases, and ensure customer can easier to find the correct path to configure flexPath.

patrick-east: My 2c, we should go with what @taherv suggested earlier in this thread where the `flexPath` is not specific to the k8s/openshift stuff. In addition we need to get rid of the `flexBaseDir` junk, it doesn't make sense to support the old images with the new helm charts anymore. IMO it just complicates this whole thing.

Then we need to have documented (in values.yaml and in the README) how someone can find what the value should be. We can provide some commented out "usual suspects" and info about when they are known to work. Think of something like:

```
orchestrator:
  # name is either 'k8s' or 'openshift'
  name: k8s
  
  # `flexPath` needs to align with kubelet "volume-plugin-dir" configuration
  # by default in Kubernetes it is <blah>
  # Select an option below or customize for the environment.
  
  # Default for Kubernetes and OpenShift 3.11+
  flexPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec
    
  # OpenShift <3.11 defaults to
  #flexPath: /whatever/ose/uses
  
  # OpenShift on Atomic (containerized kubelet/origin-node) defaults to
  #flexPath: /the/atomic/path
  
  # RKE path
  #flexPath: /some/other/weird/thing/probably

 # etc etc
```

IMO this fits in way better with how people use the helm charts. Step 1 is to look at the github readme to see the command to install it, step two is to grab the values.yaml and start editing things. If its a simple choice in values.yaml that can be customized then IMO this makes things a lot easier and would solve problems.

dinathom:
Addressed by 9871d644fc9660e671870fbe5cf6e0320dd526d1

PR #61, opened 2019-02-05
Add nodeSelector, tolerations, and affinity

japaniel:
We add taints to some of our nodes for various reasons. This will help us adjust our values.yaml to account for those. 

I did try to address these in what seems to be a commonly seen pattern for helm. I am definitely open to feedback if there is a different/prefered way to do this.


japaniel:
Thanks @patrick-east. Cut at your normal intervals will be fine for now. I was more wondering if there was a cadence that I can look at to see some indicator in terms of time frame.

PR #64, opened 2019-02-28
Typo fix

rlondner:
Fixed NFSEndPoint typo in pure-k8s-plugin/values.yaml


dinathom:
Not needed.  Both NFSEndPoint NfsEndPoint  are both acceptable as of version 2.3.0

PR #65, opened 2019-03-11
Fix flexPath in comment for Openshift 3.10+ on Atomic RHEL in values.…

dinathom:
…yaml


sdodsley:
LGTM

taherv:
Dead comment code always rots over time.

Maybe we should have different values.yaml files per environment ? Or get creative in other ways .....

dinathom:
@taherv we have not figured out a better place to document all the flexPath configs. While we think about that - this PR will just make existing comments more reliable -- so a positive to merge for now.

PR #66, opened 2019-03-11
Dinathom/nodeselect

dinathom:
Remove hard-coded nodeSelector from pure-flex template. 
By default our pure-flex will not have a nodeSelector specified which means we will default to using whatever is in the project-wide nodeSelector. If project-wide node selector is not specified it will use the cluster-wide node selector. See https://docs.openshift.com/container-platform/3.11/admin_guide/managing_projects.html#using-node-selectors

### DIFF
--- a/pure-k8s-plugin/README.md
+++ b/pure-k8s-plugin/README.md
@@ -44,7 +44,7 @@ The following table lists the configurable parameters and their default values.
 | `orchestrator.name`         | Orchestrator type, such as openshift, k8s | `k8s`                              |
 | `flexPath`                  | Full path of directory to install flex plugin, works with image.tag >= 2.0.1 | `/usr/libexec/kubernetes/kubelet-plugins/volume/exec` |
 | *`arrays`                    | Array list of all the backend FlashArrays and FlashBlades | must be set by user, see an example below                |
-| `nodeSelector`              | [NodeSelectors](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector) | `{}` |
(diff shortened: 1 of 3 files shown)


sdodsley:
Have you confirmed that removing this will pick up the defaultNodeSelector setting?

dinathom:
@sdodsley tested on Openshift 3.11 
1. in pso-namespace + no annotations on namespace. Daemonset pods get scheduled on all nodes labeled compute=true even as new compute nodes are added. 
2. If namespace has annotation openshift.io/node-selector: "" then daemonset pods are scheduled on all nodes irrespective of defaultNodeSelector in master-conf.yaml

sdodsley:
@dinathom what happens with no annotations on pso namespace and you change {{defaultNodeSelector}} to be {{compute}} and {{master}}, then install PSO? Does it deploy pods on compute and master nodes?

dinathom:
@sdodsley I have not found a 'set union' syntax with nodeSelector. According to the k8s manual nodeSelector is just a set of  key=value labels and multiple labels will mean -intersection of the labels. So I do not know how you would specify 'compute-and-master' in the master-config.yaml 
One way is to create a new label and apply it to both compute and master nodes.

sdodsley:
@dinathom I'm struggling to find where I saw it. I agree that nodeSelector is a KVP map so you can't have two keys the same. I think what I may have read is that OpenShift don't mind if you use master for user pods now. 
This is all so confusing :) 
I still think we should go with the {{defaultNodeSelector}} value and insist in our documentation that the PSO installation uses its own project and not default.
I will ensure this is covered in the OpenShift 3.11 Ref Arch I'm writing.

PR #68, opened 2019-03-15
Use the uninstaller binary

taherv:


dinathom:
We are not merging this, right? Should we close it

Issue #70, opened 2019-03-18
Only create specified storage classes.


gregsidelinger:
Provide a way to enable/disable the three different storage classes.  For example I'm not setting up pure-file however the helm chart assumes I am and creates a storage class for it.

taherv:
Hi Greg,
You never know when you might deploy a FlashBlade in your environment. Bad joke, I know, moving on ....

This might impact the usability aspect of the plugin, in a few ways
a. An extra knob on install to choose which storage classes are needed
b. Some way of going back and add the missing storage classes, if one does changes their fleet to need one.

You can always delete any of the builtin storage classes after install, if you think (b) will never happen to you. BTW, the "pure" storage class is provided for compatibility only. Feel free to delete that as well, and use pure-block for FlashArray.

gregsidelinger:
I do not agree with having to bust out kubectl to cleanup the work a helm chart did.  Also extra knobs should not matter as long as they have good defaults.

I worked out an update to the chart this morning that provides backwards compatibility with existing settings but will allow each storageclass to be managed own it's own going forward.

sdodsley:
This is no longer required due to latest updates and additional SC documentation

PR #71, opened 2019-03-18
Storageclass updates

gregsidelinger:
https://github.com/purestorage/helm-charts/issues/70

I broke out each storage class into their own section. This way other settings can be tweaked as the chart and driver progress.  I exposed an enabled and default settings to start with as well as the storage backend for pure.  storageclass.pure will use the old storageclass.isPureDefault and storageclass.pureBackend if set for backwards compatibility.


sdodsley:
Another thought. If `helm upgrade` is run and the default SC has been changed then this will cause there to be 2 default SCs which k8s will take to mean no default. 
The Chart should attempt to check the existing default and if changed remove the original default.

gregsidelinger:
I don't think helm supports looking at old values during an upgrade. And at some point the kube admins need to make sure they are not trying to set more than one default across the different storage providers, not just the pure ones.

I understand the pure backend being there for backwards compatibility as I think removing in use storage classes is dangerous.  This approach would allow new deployments to not create the old/un-used storage classes.  Maybe some good comments and documentation would help new deployments know what backends to enable.

This is only one approach.  If you have better ideas that do not involve running `kubectl delete storageclass` after every helm upgrade have at it.

sdodsley:
I'm not saying remove the storageClass `pure` just do not reference it in `values.yaml`.

pure-k8s-plugin/templates/storageclass.yaml
```diff
@@ -1,34 +1,44 @@
+{{- if .Values.storageclass.pure.enabled -}}
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
   name: pure
   annotations:
-    storageclass.beta.kubernetes.io/is-default-class: "{{ .Values.storageclass.isPureDefault }}"
```
taherv: Is this change required ? The default is already set to false, in values.yaml ...

pure-k8s-plugin/values.yaml
```diff
@@ -14,10 +14,28 @@ app:
 
 # do you want to set pure as the default storageclass?
 storageclass:
-  isPureDefault: false
+  # These are here for backwards compatibility only.
+  # New deployments should not use these settings
+  # isPureDefault: false
```
taherv: Why is this commented out ? Don't we want isPureDefault's value to be false like in previous versions ?

pure-k8s-plugin/values.yaml
```diff
@@ -14,10 +14,28 @@ app:
...
   # pureBackend: file
 
+  # This is here for backwards compatibility only.
+  # New deployments should set storageclass.pure.enabled to false
+  pure:
+    enabled: true
+
+  pureFile:
+    enabled: true
+    default: false
```
taherv: I wouldn't do this, because it is easy to have two defaults (e.g if somebody sets pureFile.default = true and pureBlock.default = true in values.yaml

pure-k8s-plugin/values.yaml
```diff
@@ -14,10 +14,28 @@ app:
...
+
+  pureBlock:
+    enabled: true
+    default: false
+
+backwards:
```
taherv: What is this "backwards:" for ?

taherv:
Hi @gregsidelinger 
I provided some code review feedback for you.

I think I still have concerns around merging it though. 

These storage classes are "starter" classes to get a vanilla system configured for maximum compatibility, and ease of install. The starter storage classes will grow in the future (think pure-replicated, and pure-cloud-backup-enabled etc.), and adding knobs per storageclass will not scale.
We (Pure, and Kubernetes in general, know that Kubernetes admins will want to tailor storage classes to their environment). 

A suggestion, design a customization step in your environment and automate/implement those storage classes along with your helm install step in your automation.

taherv:
@gregsidelinger Thinking a little out of the box here, how about an approach where we provide builtin storageclasses or none at all (controlled by a flag) ? That way there will be only one knob, if you want the inbox ones, or you want to create yours from scratch (of course there will be documentation/guidance).

patrick-east:
+1 for the all or nothing approach on the storage classes. A long time ago we added these to our installer as more like examples and they have (for better or worse) become kind of canonical things. IMO they really should only be for like POC/getting started helpers and deployers should use more customized StorageClasses to match their needs.

Just thinking out loud, but If we want to have more customizable ones we could just let them be declared in the values.yaml and we shim that yaml into sc templates.

Something (super high level) like:

```
{{ if .Values.createExampleStorageClasses }}
// The junk we have now
{{ end }}
{{ range .values.additionalStorageClasses }}
// drop the yaml from values into the sc metadata with the other boilerplate in this template
{{ endif }}
```
and a values.yaml could have some section like:
```
additionalStorageClasses:
  - name: my-default-sc
    annotations:
      storageclass.beta.kubernetes.io/is-default-class: "false"
    labels:
      kubernetes.io/cluster-service: "true"
  - name: some-other-sc
    annotations:
      storageclass.beta.kubernetes.io/is-default-class: "true"
    labels:
      kubernetes.io/cluster-service: "true"
```

We could even just define the pure, pure-file, and pure-block ones that way and drop the static definitions for them in the templates altogether.

Again, just thinking out loud. Am curious to hear thoughts from @gregsidelinger , @taherv, @sdodsley, and @dinathom

pure-k8s-plugin/values.yaml
gregsidelinger: That should not have been there, that's what happens work while on vacation.

pure-k8s-plugin/values.yaml
gregsidelinger: It's a side affect of having three storage classes in the same helm chart. I'm not sure of a good way to have helm yell at you for marking more than one default.  An alternative might be to have an option like so.

```
storageclass:
  # default: pureBlock
  # default: pureFile
  default: 
```
And then do something like this in in the pureBlock storage class template.
```
{{ if eq .Values.storageclass.default  eq "pureBlock" }}
  annotations:
    storageclass.beta.kubernetes.io/is-default-class: "true"
{{ end }}
``

pure-k8s-plugin/templates/storageclass.yaml
gregsidelinger: An IF statement could be used  to only set the annotation when it is true.  The alternative "default" option I mentioned below would use IF statements.

gregsidelinger:
I was thinking about something like your additionalStorageClasses at first but figured it was more work.  Loops in helm are no fun if you want to access something in the _helpers.tlp since most functions contain references to `.` which is is the local loop var and not what they expect.  That would allow users to pick different names and maybe different reclaimPolicy per class so it has my +1.

sdodsley:
This is no longer required due to latest updates and additional SC documentation

PR #73, opened 2019-03-20
Separate nodeSelector, tolerations and affinity config for flexDaemon and provisioner

javefang:
These should be separate settings for flexDaemon and provisioner.

In a cluster with tainted nodes, for pure driver to work on all nodes, it's necessary to add tolerations for the flexDaemon to run on all nodes, but it doesn't necessarily mean the provisioner should have the same tolerations (or nodeSelector/affinity).


taherv:
Thanks for your contribution !

You might want to do the update, so that the tarball for 2.4.0 is generated correctly.
Refer to "Updating Helm Charts" on the top level README.md.

sdodsley:
Let's bump the version number to 2.3.1 here

javefang:
I reverted the version bump commit, since I'm not sure what's the release process here. Maybe worth leaving the version bump to you?

javefang:
This could be counted as a breaking change as well, as some options has been removed (renamed)

taherv:
Good catch.
Perhaps add in the new ones, instead of replacing them ?
We will deprecate the older ones, it might take a few releases before we just delete them.

Also, @javefang, updating/releasing has to happen in the same commit, otherwise an install using 2.3.1 will fail if it doesn't find the tarball. Its not very tedious, just run update.sh and add in the tarball to your Pull Request.

taherv:
Oh, I didnt read your comment about version bump.

You are right. Please update the fields (add vs replace), we are then good to merge.

javefang:
Ok, I didn't see the tarballs are in the repo as well. 

So I will

1. Update chart version to 2.3.1
2. Run update.sh
3. Commit everything and push

Is that correct?

pure-k8s-plugin/README.md
```diff
@@ -45,9 +45,12 @@ The following table lists the configurable parameters and their default values.
 | `orchestrator.name`         | Orchestrator type, such as openshift, k8s | `k8s`                              |
 | `flexPath`                  | Full path of directory to install flex plugin, works with image.tag >= 2.0.1 | `/usr/libexec/kubernetes/kubelet-plugins/volume/exec` |
 | *`arrays`                    | Array list of all the backend FlashArrays and FlashBlades | must be set by user, see an example below                |
-| `nodeSelector`              | [NodeSelectors](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector) Select node-labels to schedule flex-plugin. See [this](https://docs.openshift.com/container-platform/3.11/admin_guide/managing_projects.html#using-node-selectors) for setting node selectors on Openshift. | `{}` |
```
taherv: The review feedback was to keep/retain nodeSelector, toleratons and affinity, and add in flexDaemon.nodeSelector and the other options ..... this helps compatibility. Could you please add these back in ?
If you see issues with that, please highlight them.

javefang: nodeSelector/tolerations/affinity is introduced in the recent release (2.3.0), and to me it is confusing not to have separate keys for flex and provisioner. Is it worth keeping it for backward compatibility? 

If the answer is yes, then I guess it will work like this:

- if `nodeSelector` is set, it override both `flexDaemon.nodeSelector` and `provisioner.nodeSelector`
- if `nodeSelector` is not set, it uses value from the other two

Same applies for `tolerations` and `affinity`.

How does that sound?

taherv: That would work !

taherv: We can follow up with some more documentation (you dont have to do it) about best practices of using the new values added by you.

javefang: EDIT, it's probably better to prefer `flexDaemon.nodeSelector` over `nodeSelector` if both are set. I'll make the change and update the doc.

javefang: ok done! please let me know if this looks good, then I'll generate the tarball

pure-k8s-plugin/pure-man-dev.yaml
```diff
@@ -0,0 +1,26 @@
+arrays:
```
taherv: You probably dont want to check this in ....

pure-k8s-plugin/templates/pure-flex-daemon.yaml
```diff
@@ -73,15 +73,15 @@ spec:
         - name: kubelet-plugins
           hostPath:
            path: {{ template "pure_k8s_plugin.flexpath" .}}
-    {{- with .Values.nodeSelector }}
+    {{- with .Values.flexDaemon.nodeSelector | default .Values.nodeSelector }}
```
taherv: Nice !

pure-k8s-plugin/README.md
taherv: LGTM, except for the man-dev.yaml file.

javefang: oops...

javefang: should be gone now, and we have revoked that token

pure-k8s-plugin/pure-man-dev.yaml
javefang: (token revoked and removed from the PR)

javefang:
Hi @taherv , I have built and pushed the tarball now. Could you approve, please?

PR #75, opened 2019-03-25
Adding additionalStorageClasses option

gregsidelinger:
Adding the additionalStorageClasses option that @patrick-east suggested in https://github.com/purestorage/helm-charts/pull/71.  Three example classes have been added to the values.yaml that will produce three StorageClasses identical to the three default ones.

A storageclass.create option has also been added that allows uses to turn off the creation of the default three StorageClasses. The default value is true to maintain backwards compatibility.


sdodsley:
This is no longer required due to latest updates and additional SC documentation

PR #76, opened 2019-03-29
Chart icon fix and enhancements

rlondner:
- Fixed icon url
- Added `keywords` and `maintainers` attributes


pure-k8s-plugin/Chart.yaml
```diff
@@ -1,6 +1,11 @@
 apiVersion: v1
 description: A Helm chart for Pure Storage persistent storage driver for Kubernetes/OpenShift
 name: pure-k8s-plugin
-version: 2.3.1
+version: 2.3.2
```
sdodsley: I don't think we should be changing these release numbers in PRs. This should be down to the release process to pick the correct release number this PR goes into.

PR #84, opened 2019-04-25
2.4.1 Release

dinathom:
This release updates helm-chart to use k8s:2.4.1 image
We have also made PSO Operator usage more friendly.
- hostpath SCC is now created and added to the service account of the installer by teh script
- upgrade.sh script is provided to apply changes to values.yaml easily to PSO Operator
- flexPath changes are now compatible with versions prior to 2.2.1


sdodsley:
Still don't see the update script

dinathom:
> Still don't see the update script

https://github.com/purestorage/helm-charts/blob/master/operator/upgrade.sh

sdodsley:
@dinathom I think we should call this update.sh, not upgrade.sh. The file is there to help customers update their YAML file with new configs. They don't have to be upgrading...

PR #91, opened 2019-05-03
2.5.0 Release: Upgrade external storage controller lib and block retain.

pure-yesmat:


pure-yesmat:
@dinathom pointed out that I was missing the changes to install.sh for the RBAC changes. I updated the PR with those changes.

operator/install.sh
```diff
@@ -260,6 +290,26 @@ rules:
...
+  - apiGroups:
+    - rbac.authorization.k8s.io
+    resources:
+    - roles
+    - rolebindings
+    verbs:
+    - "create"
+    - "delete"
+    - "list"
+    - "get"
```
dinathom: Will need "patch" role for upgrades or applying a new values.yaml.

For Role I would suggest use "*" since it is restricted to this namespace only.

operator/install.sh
```diff
@@ -260,6 +290,26 @@ rules:
...
+  - apiGroups:
+    - authorization.openshift.io
+    resources:
+    - roles
+    - rolebindings
+    verbs:
+    - "create"
+    - "delete"
+    - "list"
```
dinathom: Same comment as above.

PR #102, opened 2019-07-10
Add CSI plugin directory

dinathom:
Adding the CSI plugin helm chart. Our CSI plugin follows spec 1.0.0  and works on k8s 1.13+
The chart will setup all the necessary sidecars for a k8s install for 1.13 , 1.14 and 1.15 (Yes, each has its own special mix of sidecars ).

Our pure-csi-plugin chart works with the values.yaml from our pure-k8s-plugin flexvolume chart  with just the new image tag.

NOTE: On k8s 1.13 CSI is still alpha and needs to be enabled using the CSIDriverRegistry feature-gate on kubelet and the control-plane.


pure-csi-plugin/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+description: A Helm chart for Pure Service Orchestrator CSI driver
+name: pure-csi-plugin
```
pure-yesmat: so the name is different and the version is 3.0, shouldn't the name be the same?

pure-csi-plugin/templates/node.yaml
```diff
@@ -0,0 +1,233 @@
+{{ if and (eq .Capabilities.KubeVersion.Major "1") (lt .Capabilities.KubeVersion.Minor "14") }}
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: csidrivers.csi.storage.k8s.io
```
pure-yesmat: So there is only one of these per cluster correct? What happens if another CSI driver is already installed will this fail?

pure-csi-plugin/templates/node.yaml
```diff
@@ -0,0 +1,233 @@
...
+            name: socket-dir
+          image: quay.io/k8scsi/livenessprobe:v1.1.0
+          args:
+          - --csi-address=/csi/csi.sock
+          - --connection-timeout=3s
+          - --health-port=9898
+
+      volumes:
+        -  secret:
+            secretName: pure-provisioner-secret
```
pure-yesmat: spacing

pure-csi-plugin/templates/node.yaml
```diff
@@ -0,0 +1,233 @@
...
+    {{- with .Values.mounter.affinity | default .Values.affinity }}
+      affinity:
+{{ toYaml . | indent 8 }}
+    {{- end }}
+    {{- with .Values.mounter.tolerations | default .Values.tolerations }}
+      tolerations:
+{{ toYaml . | indent 8 }}
+    {{- end }}
+{{- else -}}
+{{- if .Values.flexDaemon -}}
```
pure-yesmat: flexDaemon?

pure-csi-plugin/templates/rbac.yaml
```diff
@@ -0,0 +1,136 @@
...
+#   this influences the RBAC setup, see below
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.clusterrolebinding.serviceAccount.name }}
+  namespace: {{ .Release.Namespace }}
+
+---
+kind: ClusterRole
```
pure-yesmat: we don't need any roles/rolebindings? for the endpoints resources?

pure-csi-plugin/templates/node.yaml
dinathom: We want to ensure that the values.yaml that was used with our flex-driver can be used with CSI.

operator-csi-plugin/README.md
```diff
@@ -0,0 +1,70 @@
...
+## Installation
+
+A single install script sets up the Pure CSI Operator. <br/>
+```install.sh --image=<image> --namespace=<namespace> --orchestrator=<ochestrator> -f <values.yaml>```
+
+Parameter list:<br/>
+1. ``image`` is the Pure CSI Operator image. If unspecified ``image`` resolves to the released version at [quay.io/purestorage/pso-operator](https://quay.io/purestorage/pso-operator).
+2. ``namespace`` is the namespace/project in which the Pure CSI Operator and its entities will be installed. If unspecified, the operator creates and installs in  the ``pure-csi-operator`` namespace.
+**Pure CSI Operator MUST be installed in a new project with no other pods. Otherwise an uninstall may delete pods that are not related to the Pure CSI Operator.**
+3. ``orchestrator`` defaults to ``k8s`` since ``openshift`` does not have CSI v1.0.0 support yet.
```
sdodsley: OpenShift 4.1 does have CSI support, but it looks like it's 1.0-rc1 so, in theory, our CSI driver should still work in OpenShift 4.1+

operator-csi-plugin/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -xe
+PSO_OPERATOR_IMAGE_TAG=${PSO_OPERATOR_IMAGE_TAG:-quay.io/dinathom/my-pso-operator:csi-v1.0.0}
```
sdodsley: this should be a Pure location, not yours...

operator-k8s-plugin/README.md
```diff
@@ -1,26 +1,26 @@
-# PSO Operator
+# Pure Flex Operator
 
 ## Overview
-PSO Operator is now  the preferred install method for PSO on OpenShift 3.11 and higher versions. 
-The PSO Operator packages and deploys the Pure Service Orchestrator (PSO) on OpenShift for dynamic provisioning of persistent volumes on FlashArrays and FlashBlades. The minimum supported version is OpenShift 3.11.
+Pure Flex Operator is now  the preferred install method for PSO on OpenShift 3.11 and higher versions.
```
sdodsley: check your spacing

pure-csi-plugin/README.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+## How to upgrade from the flexvolume to CSI
+
+Upgrade from flexvolume to CSI is not currently supported and will be worked on in an upcoming release.
+
+## Platform and Software Dependencies
+- #### Operating Systems Supported*:
+  - CentOS 7
+  - CoreOS (Ladybug 1298.6.0 and above)
```
sdodsley: We need to be checking against Red Hat Enterprise CoreOS 4.1 as this is the required OS for OpenShift 4.1+

pure-csi-plugin/templates/node.yaml
sdodsley: Good question... It is very possible a customer would want more than 1 CSI driver in their cluster to provide different tiers of storage

pure-csi-plugin/templates/storageclass.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: pure
```
sdodsley: Please, please, please let's get rid of the pure SC and have a setting to define if pure-block or pure-file is the default SC. TBH I would like to stop providing any SCs in the installation and just document how to create on for FA or FB. Customers are starting to think these are the only SCs we support

pure-csi-plugin/Chart.yaml
dinathom: These are different charts. There is no relation between pure-k8s-plugin-2.5.2.tgz and pure-csi-plugin-3.0.0.tgz 
So the name should be different.  We could chose any number in the version and the reason to chose 3.0.0 was to make this follow the image tag

operator-csi-plugin/README.md
dinathom: I'm going to drop this line from the README and yes we should work with whichever version has CSI 1.0 support

pure-csi-plugin/templates/rbac.yaml
dinathom: These rights are copied  from the external-provisioner repo 
https://github.com/kubernetes-csi/external-provisioner/tree/master/deploy/kubernetes

pure-csi-plugin/templates/node.yaml
dinathom: Well then the right thing is to remove this from the chart and assume (correctly) that the cluster has been setup to handle CSI.
So customers trying our chart on 1.13 will have to create this CRD as mentioned in the documentation: https://kubernetes-csi.github.io/docs/csi-driver-object.html

sdodsley: Although that page does say you have to have this object, but that you can list drivers, plural, so that implies there can be multiple that use the same object... Maybe we should test this before making a decision on what to do here.

taherv: We cannot assume that the CRD will be present. As far as a customer is concerned, the first CR should bring in the CRD (or we have to write an operator, but that is overkill).
At most we can try to create the CRD, and if it exists, treat it as not-an-error and proceed.

I advise against asking the customer to create the CRD ...

dinathom: Not sure I understand your point. 
We are testing this in our CI/CD - this chart does work for one driver and fail for multiple drivers.

dinathom: I didnt find a easy way in helm to ignore errors.

taherv: Hi, One last question .... How is this working ?
a. Chart install installs CRD.
b. Chart install installs CR for Pure's CSI Driver object.

I thought (b) could fail due to a timing issue, that the CRD's resource has not been completely initialized in the api server.

dinathom: On 1.13 
1. Chart install - installs CRD no CR is installed yet. CR is created by the sidecar container cluster-driver-registrar after the installation is complete.

pure-yesmat: We can have a separate CRD.yaml we provide and have documentation that if the CRD is not installed they should run this before running the helm install.

Spending too much time to make this seemless is not worth it in my opinion. I am guessing that this will be part of k8s in later releases anyways.

pure-csi-plugin/templates/node.yaml
pure-yesmat: I agree, but this might be confusing to some customers.

pure-csi-plugin/templates/rbac.yaml
pure-yesmat: Where are they installed/applied though?

pure-csi-plugin/templates/node.yaml
dinathom: We could just add this to the README: I agree that this is getting to be a time sink on k8s 1.13
This issue does not exist on 1.14 and higher versions since it is a builtin once CSI became beta.

For K8s 1.13 ensure that CSIDriver is enabled by
1. Add  "--feature-gates=CSIDriverRegistry=true" to kubectl and the control-plane. 
2. Create the CSIDriver CRD by running the following:
 $> kubectl create -f https://raw.githubusercontent.com/kubernetes/csi-api/master/pkg/crd/manifests/csidriver.yaml

pure-csi-plugin/templates/node.yaml
dinathom: How about I add this line to the comment:
# Look at the flexDaemon nodeSelector values too because we may be using a flex-plugin values.yaml and we honor the old values.yaml

pure-csi-plugin/templates/rbac.yaml
dinathom: ClusterRoles/RoleBindings are granted via service accounts. Example see provisioner.yaml
here the 
      serviceAccountName: {{ .Values.clusterrolebinding.serviceAccount.name }}
ensures that the sidecar below has enough rights.
  - name: csi-provisioner
          image: quay.io/k8scsi/csi-provisioner:v1.1.0

Does this answer the question?

pure-csi-plugin/templates/storageclass.yaml
dinathom: This is a separate issue and I do not want to mix it with the CSI driver release. Can we create a new issue for this and it will be addressed after the CSI release.

pure-csi-plugin/templates/node.yaml
pure-yesmat: I agree

pure-csi-plugin/templates/node.yaml
pure-yesmat: Maybe @sdodsley can comment. I don't have a strong opinion.

pure-csi-plugin/templates/rbac.yaml
pure-yesmat: So I understand the cluster roles. For the roles and rolebindings you pointed to the external-provisioner. But how do those roles get installed?

dinathom: Ah ok, no Role/Rolebindings are specified for these sidecars.

pure-csi-plugin/templates/node.yaml
dinathom: I was able to write a job to run and test if there is no-CRD only then create this. For this release I am only using the README. Will include the job in the helm-chart as a pre-install hook in a follow-on PR

pure-csi-plugin/templates/node.yaml
sdodsley: I think your comment is fair. We do want to honour old values, particularly when we come up with an upgrade process.

sdodsley:
Can someone tell me the name of the CSI plugin - apparently there is an official dot notation naming convention. I need this to be able to add our driver into the k8s CSI doc set.

dinathom:
> Can someone tell me the name of the CSI plugin - apparently there is an official dot notation naming convention. I need this to be able to add our driver into the k8s CSI doc set.


According to the spec here: https://github.com/container-storage-interface/spec/blob/master/spec.md
See the description under: 
```message GetPluginInfoResponse {
  // The name MUST follow domain name notation format
  // (https://tools.ietf.org/html/rfc1035#section-2.3.1). It SHOULD
  // include the plugin's host company name and the plugin name,
  // to minimize the possibility of collisions. It MUST be 63
  // characters or less, beginning and ending with an alphanumeric
  // character ([a-z0-9A-Z]) with dashes (-), dots (.), and
  // alphanumerics between. This field is REQUIRED.
```

Checking the definition under https://tools.ietf.org/html/rfc1035#section-2.3.1 , the string "pure-csi" does satisfy the requirements. 
While a lot of companies have used reverse DNS notation for their CSI plugin having a simple, easy to remember notation seems like the best choice.

sdodsley:
Whilst i agree simple is better there are comments in other PRs requesting the dot notation over a simple string

taherv:
We know about the naming requirements from the spec, and deliberately chosen "pure-csi". The tradeoff was between ease of support and compliance with the spec, and our current decision gets both.

If you think it still needs to change, submit an RFE with your reasoning :)

dinathom:
> Whilst i agree simple is better there are comments in other PRs requesting the dot notation over a simple string

Can you point to some such PRs? Where are the requests coming from -- customer or community? 
I would have thought that customers want simple,easy to remember names.
If it is the community then it seems there should be a stricter spec somewhere.

sdodsley:
Example of a name request to be in dot notation can from Saad Ali (no one important...only the Storage SIG lead) in https://github.com/kubernetes-csi/docs/pull/162 although he did say this wasn't a blocker to that specific PR

PR #106, opened 2019-07-30
Renaming pure-csi-plugin to pure-csi.

pure-yesmat:
The CSIDriver CRD name is pure-csi, this change is to match our chart name to that.


sdodsley:
I guess this means I need to change the k8s CSI docs page...

README.md
```diff
@@ -28,7 +28,7 @@ For installation, see the [Flex Operator Documentation](./operator-k8s-plugin/RE
 
 ### Helm Chart
 **pure-k8s-plugin** deploys PSO Flexvolume plugin on your Kubernetes cluster. 
-**pure-csi-plugin** deploys PSO CSI plugin on your Kubernetes cluster. 
+**pure-csi** deploys PSO CSI plugin on your Kubernetes cluster.
```
trevormeiss: If we are removing "plugin" from the name, might as well remove it from the description too?

sdodsley:
We did go through this before. Seems like this is a more standard way and the current `plugin` names are older. I was an advocate of the dot notation name, but was shot down :)
Anyway, I've raised a PR in the CSI driver docs repo for this change now, so let's not change it again...

pure-yesmat:
We chatted about this through email. The CSIDriver CRD instance was pure-csi, but our helm chart was pure-csi-plugin. We just wanted to match the two.

README.md
pure-yesmat: I don't really think its related. This is about matching the name with our CRD name. At the end its a plugin ...

sdodsley:
The rename of the driver in the CSI docs is now live: https://github.com/kubernetes-csi/docs/pull/188

README.md
trevormeiss: Got it

PR #109, opened 2019-08-06
Fix some doc errors and some logic in the Operator install scripts and add missing file

sdodsley:
Clarify details for installation of the Operator.
Fix the install script that was not telling you if your specified configuration file didn't exist.
Add the default `values.yaml` file for CSI into the `operator-csi-plugin` directory.

### DIFF
--- a/operator-csi-plugin/install.sh
+++ b/operator-csi-plugin/install.sh
@@ -2,15 +2,16 @@
 IMAGE=quay.io/purestorage/pso-operator:v5.0.0
 NAMESPACE=pure-csi-operator
 KUBECTL=kubectl
+ORCHESTRATOR=k8s
 
 usage()
 {
     echo "Usage : $0 --image=<imagename> --namespace=<namespace> --orchestrator=<orchestrator> -f <values.yaml>"
-    exit
 }
 
 if [[ "$1" == "-h" || "$1" == "--help" ]]; then
     usage
+    exit
 fi
 
 while (("$#")); do
@@ -53,8 +54,8 @@ case "$1" in
 done
 
 if [[ -z ${VALUESFILE} || ! -f ${VALUESFILE} ]]; then
+    echo "File ${VALUESFILE} does not exist"
     usage
-    echo "File ${VALUESFILE} for values.yaml does not exist"
     exit 1
 fi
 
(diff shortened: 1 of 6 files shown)


operator-csi-plugin/install.sh
```diff
@@ -53,8 +54,7 @@ case "$1" in
 done
 
 if [[ -z ${VALUESFILE} || ! -f ${VALUESFILE} ]]; then
-    usage
-    echo "File ${VALUESFILE} for values.yaml does not exist"
+    echo "File ${VALUESFILE} does not exist"
```
pure-yesmat: I would actually say we should just keep it the way it was and take out the exit line in the usage() func.

operator-k8s-plugin/install.sh
```diff
@@ -53,8 +54,7 @@ case "$1" in
 done
 
 if [[ -z ${VALUESFILE} || ! -f ${VALUESFILE} ]]; then
-    usage
-    echo "File ${VALUESFILE} for values.yaml does not exist"
+    echo "File ${VALUESFILE} does not exist"
```
pure-yesmat: same here

operator-csi-plugin/install.sh
sdodsley: I removed the values.yaml reference because if they use the file values.yaml the error message is just weird :)

operator-k8s-plugin/install.sh
sdodsley: As above

operator-csi-plugin/install.sh
pure-yesmat: oh yeah that makes sense.
I just meant we should keep the usage followed by error and git rid of the exit line in the usage function

sdodsley: can't remove exit from the usage function as there is a point it is called without an exit immediately after

sdodsley: You don't really need the usage call when all that is wrong is the filename. Therefore is nothing else wrong with the script call or it would have been invoked earlier.

pure-yesmat: the only place is line 14, right? Can you add it there since you are editing the file.

PR #111, opened 2019-08-14
allow csi images to be customisable

javefang:
allow all quay.io/k8scsi/* images to be customisable. necessary for those k8s clusters that don't have access to public docker registries.

(I haven't updated `Chart.yaml` and hasn't run`update.sh` yet. Could someone advise where can I find the doc for the release process, please?)

### DIFF
--- a/pure-csi/README.md
+++ b/pure-csi/README.md
@@ -80,16 +80,12 @@ The following table lists the configurable parameters and their default values.
 | `provisioner.tolerations`               | [Tolerations](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#concepts)  | `[]` |
 | `provisioner.affinity`                  | [Affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) | `{}` |
 | `csi.provisioner.image.name` | The image name of the csi-provisioner | `quay.io/k8scsi/csi-provisioner` |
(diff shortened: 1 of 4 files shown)


pure-yesmat:
Interesting.

I understand the need to change the location of the images, but changing the tag could be problematic because our driver could depend on features in a specific version/tag.

Perhaps we should just have a customizable registry, ie prefix to the image name and tag. Something like:

```yaml
image: "{{.Values.image.registry}}k8scsi/csi-provisioner:v1.1.0"
```

javefang:
Hi @pure-yesmat , it's quite common in helm charts where image could be customised. Your concern about tag does make sense, however I'm not exactly sure how other users are going to tag their image so it's probably better to leave it fully customisable.

Since image name and tag are already separated and have reasonable default set, there is very little chance anyone could break it while overriding the image name. The only way they can break it is when then override the tag, which they have been warned.

pure-yesmat:
sorry for the delay @javefang, i'll get back to you by the EOD.

pure-yesmat:
Hi @javefang,

I think we are ok with having a customizable name but not tag. So we can have it something like:
```yaml
image: "{{.Values.image.name}}:v1.1.0"
```

javefang:
Hi @pure-yesmat, I have udpated my PR to disallow csi image tags to be customised.

Pure-AdamuKaapan:
Tested locally and seems to apply values as expected (mainly by setting them to purposefully wrong values and verifying that they crash as expected, since I don't have those images pushed to a repo of our own and figured this was a faster test).

PR #115, opened 2019-08-30
Add feature-gate support :VolumePVCDataSource,VolumeSnapshotDataSource

dsupure:
Add feature gate support for enabling CSI 1.0 clone and snapshot.

### DIFF
--- a/pure-csi/templates/provisioner.yaml
+++ b/pure-csi/templates/provisioner.yaml
@@ -68,14 +68,28 @@ spec:
 
         # This is the external provisioner sidecar
         - name: csi-provisioner
+{{- if eq .Values.featuregate.VolumePVCDataSource true}}
+          image: quay.io/k8scsi/csi-provisioner:v1.4.0-rc1
+{{ else }}
           image: quay.io/k8scsi/csi-provisioner:v1.1.0
+{{- end }}
           args:
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
-
+{{ if eq .Values.featuregate.VolumeSnapshotDataSource true}}
+        - name: csi-snapshotter
+          image: quay.io/k8scsi/csi-snapshotter:v1.1.0
+          args:
+            - "--csi-address=/csi/csi.sock"
+            - "--connection-timeout=15s"
+            - "--leader-election=false"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+{{ end }}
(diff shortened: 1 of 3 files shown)


pure-csi/templates/provisioner.yaml
```diff
@@ -68,14 +68,28 @@ spec:
 
         # This is the external provisioner sidecar
         - name: csi-provisioner
+{{- if eq .Values.featuregate.VolumePVCDataSource true}}
+          image: quay.io/k8scsi/csi-provisioner:v1.4.0-rc1
```
sdodsley: do we want to use `latest` as the tag rather than a specific version?

pure-csi/templates/provisioner.yaml
```diff
@@ -68,14 +68,28 @@ spec:
...
           args:
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
-
+{{ if eq .Values.featuregate.VolumeSnapshotDataSource true}}
+        - name: csi-snapshotter
+          image: quay.io/k8scsi/csi-snapshotter:v1.1.0
```
sdodsley: same here...

pure-csi/templates/provisioner.yaml
sdodsley: do we want to use `latest` as the tag rather than a specific version?

pure-csi/templates/provisioner.yaml
pure-yesmat: I prefer to target versions specifically. Sometimes the next version has breaking changes. Also it will match what we are testing in house.

sdodsley: i understand that, but we cannot be left in the same position as we currently are with the `helm-operator` image, ie a really old version with bugs and missing the latest security patches. I behoves us to ensure we ship the most recent and most stable code we can for our customer base. 
Maybe we need a monthly review of the images we use to see what has changed and check the latest out in our CI system.

dsupure: I think monthly review to make sure we have tested the images in our CI system is good.  That we can make sure that our target versions are close to up to date ones.

sdodsley:
Seems to conflict with #125. Can we resolve the conflict or close this PRE down?

dsupure:
> Seems to conflict with #125. Can we resolve the conflict or close this PRE down?

I'll close this one. 
I already put the VolumePVCDataSource and  VolumeSnapshotDataSource feature-gate explanation in #124. Customers can use ".Values.csi.snapshotter.image" to decide if they want to install external snapshotter or not. It should be good enough.

Issue #114, opened 2019-08-23
pure-csi not found in the list of registered CSI drivers


czankel:
I'm getting the following error trying to mount a file system:

`csi_attacher.go:330] kubernetes.io/csi: attacher.MountDevice failed to create newCsiDriverClient: driver name pure-csi not found in the list of registered CSI drivers`

However, the driver shows up under csidrivers:

`$ oc get csidrivers.storage.k8s.io
NAME       CREATED AT
pure-csi   2019-08-22T17:00:13Z`

Note that this is for FlashBlade, using the 'operator-csi-plugin' and some snapshot of Kubernetes v1.14.0.

sdodsley:
You say k8s 1.14, but you are also using `oc` as the command. Can we assume you are using a version of OpenShift 4.x? What version would that be?

sdodsley:
We are probably going to need a lot of log files to work out where the issue lies.
Can you post the `describe` from the pod that is failing the mount, as well as the logs from the `pso-operator` and `pure-provisoner` pods and the `pure-csi` pod from the node that is running the failing application pod. System log files from that node would also help.

pure-yesmat:
The logs @sdodsley pointed out will be important.
Can you also check if this path is created on the node: /var/lib/kubelet/plugins/pure-csi/csi.sock.

czankel:
I found the issue. but don't really have a true fix but a work-around. This probably only affects openshift that is more pedantic with security: The service account _default_ didn't have enough privileges. 

1. Add the 'privileged' security context to the default account:
`oc adm policy add-scc-to-user privileged system:serviceaccount:pure-csi-operator:default`

2. Force a restart of the pure-csi daemonset (otherwise it might take a while to take effect)
`oc delete daemonset pure-csi`

Not sure if there's a better RBAC rule to use here, but this worked.

Side-note, the service account is defined by the pso-operator:
`oc get pod pso-operator-XXXXXXXXX-YYYYY -o yaml|grep serviceAccount`
` serviceAccount: default`
` serviceAccountName: default`

pure-yesmat:
Thanks for letting us know. We were limiting the SCC to hostpath and that was sufficient for non CSI drivers pre 4.X. It seems like we may need privileged for CSI drivers in 4.X.

pure-yesmat:
@czankel I submitted PR https://github.com/purestorage/helm-charts/pull/116. Do you think you can give this a try?

czankel:
[...]

PR #116, opened 2019-09-06
Adding the privileged SCC to the CSI operator.

pure-yesmat:
According to https://github.com/purestorage/helm-charts/issues/114#issuecomment-528993222 we need privileged SCC for the CSI operator to properly install the CSI drivers on OpenShift 4.X.

It seems like this is also mentioned here: https://docs.openshift.com/container-platform/3.10/install_config/persistent_storage/persistent_storage_csi.html


sdodsley:
LGTM

pure-yesmat:
This is actually not needed. The privileged SCC needs to be added to the default user which is used by the pso-operator.

PR #123, opened 2019-10-01
CSI release 5.0.1

dsupure:
Fix logrus logger deadlock issue
Fix docker-plugin crash


sdodsley:
I think you need to update every reference to 5.0.0 in this repo. It occurs in lots of places.
Also, I don't see the 5.0.1 image in quay.io
Should probably have a release tag as well

dsupure:
@pure-yesmat and I discussed it and he thought changing the value.yaml should be good enough since helm-chart only read the values from this yaml.  However, you are right,  we should have the 5.0.1 release tag  in quay.io.  I'll do it later today.

pure-yesmat:
just to clarify, we don't need to tag in quay.io we just need to update the documentation here: https://quay.io/repository/purestorage/pso-operator

quay only has the operator and there was no operator change here.

PR #125, opened 2019-10-23
Enable CSI snapshot and clone feature

dsupure:
Enable CSI snapshot and clone features for the coming CSI 1.0.1 release.

### DIFF
--- a/pure-csi/templates/provisioner.yaml
+++ b/pure-csi/templates/provisioner.yaml
@@ -69,7 +69,7 @@ spec:
         # This is the external provisioner sidecar
         - name: csi-provisioner
           {{- with .Values.csi.provisioner.image }}
-          image: "{{ .name }}:v1.1.0"
+          image: "{{ .name }}:v1.4.0-rc1"
           imagePullPolicy: {{ .pullPolicy }}
           {{- end }}
           args:
@@ -78,7 +78,17 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
-
+        - name: csi-snapshotter
+          {{- with .Values.csi.snapshotter.image }}
+          image: "{{ .name }}:v1.1.0"
+          {{- end }}
+          args:
+            - "--csi-address=/csi/csi.sock"
+            - "--connection-timeout=15s"
+            - "--leader-election=false"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
 # This is the cluster-driver-registrar sidecar that allows helm-install without CRD-hooks for the CSIDriver CRD
 # The reason we do not want a crd-hook with helm-chart is to avoid upgrade issues like: https://github.com/helm/helm/issues/4489
 {{ if and (eq .Capabilities.KubeVersion.Major "1") (eq .Capabilities.KubeVersion.Minor "13") }}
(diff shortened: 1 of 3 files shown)


sdodsley:
I see you are using an rc1 image. Shouldnt we wait until the GA version is ready

dsupure:
> I see you are using an rc1 image. Shouldnt we wait until the GA version is ready

Looks like they have 1.4 GA. Let me do a quick test.
https://quay.io/repository/k8scsi/csi-provisioner?tag=latest&tab=tags

dsupure:
> > I see you are using an rc1 image. Shouldnt we wait until the GA version is ready
> 
> Looks like they have 1.4 GA. Let me do a quick test.
> https://quay.io/repository/k8scsi/csi-provisioner?tag=latest&tab=tags

I've verified 1.4 version works. Replace 1.4-rc1 to 1.4.

pure-csi/templates/provisioner.yaml
```diff
@@ -78,7 +78,17 @@ spec:
...
               mountPath: /csi
-
+        - name: csi-snapshotter
+          {{- with .Values.csi.snapshotter.image }}
+          image: "{{ .name }}:v1.1.0"
+          {{- end }}
+          args:
+            - "--csi-address=/csi/csi.sock"
+            - "--connection-timeout=15s"
+            - "--leader-election=false"
```
trevormeiss: [question] why don't we use leader election?

pure-csi/templates/rbac.yaml
```diff
@@ -44,16 +44,25 @@ rules:
...
     verbs: ["get", "list", "watch"]
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["create", "list", "watch", "delete", "get", "update"]
```
trevormeiss: why do we need these permissions for CRDs?

pure-csi/templates/provisioner.yaml
dsupure: The setting is suggested by the external-snapshotter yaml [example](https://github.com/kubernetes-csi/external-snapshotter/blob/master/deploy/kubernetes/setup-csi-snapshotter.yaml).

pure-csi/templates/rbac.yaml
dsupure: Same suggestion from the external-snapshotter deployment rbac [example](https://github.com/kubernetes-csi/external-snapshotter/blob/c933d6dacee9f7dc224c701c7579643c8ac700a5/deploy/kubernetes/rbac.yaml).

trevormeiss: I guess CSI needs to create VolumeSnapshotClass, VolumeSnapshot, and VolumeSnapshotContent crds:
https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/csi-snapshotter/create_crd.go

dsupure: Yeah, you are right. Thanks @trevormeiss !

PR #131, opened 2019-11-11
Add VolumeSnapshotClass to the helm-chart

dsupure:
Add VolumeSnapshotClass to the CSI helm-chart


pure-csi/templates/snapshotclass.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: snapshot.storage.k8s.io/v1alpha1
```
pure-yesmat: So I thought @sdodsley wanted to avoid adding more "classes" by default to the helm chart?

sdodsley: I wanted to avoid storageClasses as customers were getting confused by them. I think added the snapshot class is a valid thing as there will only be one of those for us I would imagine

pure-yesmat: Today there is only one. But this will be a place for customizations later I believe.

PR #134, opened 2019-11-14
Update docs to add EKS support

sdodsley:
Add AWS EKS 1.14 support for the CSI driver - both through operator and helm install methods

### DIFF
--- a/operator-csi-plugin/README.md
+++ b/operator-csi-plugin/README.md
@@ -13,6 +13,7 @@ This installation process does not require Helm installation.
   - Ubuntu 16.04
   - Ubuntu 18.04
 - #### Environments Supported*:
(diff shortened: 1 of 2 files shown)


trevormeiss:
Looks like EKS requires some initial setup of iscsi and multipath. Should we also add documentation on how to do this setup?

sdodsley:
Yes. I will put something together about that in a separate document. This is just to add it to the supported list

dsupure:
Should we add the scripts to help users to set up the iscai and multipath for EKS? We Have some handy scripts in the nsm-dev-tools.

sdodsley:
That might be a really helpful option. We can reference them in the readme and create an EKS help directory in this repo as part of the PR to add the docs

PR #136, opened 2019-11-15
Security fixes for CVE-2019-11255

pure-yesmat:
This commit is to address security concerns with presented here: https://issue.k8s.io/85233
The main fix is to upgrade csi-snapshotter to v1.2.2 (has fix) from v1.1.0 (does not have fix).

Also changed the csi image name options in the values file to optional to allow easy upgrade from before we had those options.


sdodsley:
LGTM

PR #137, opened 2019-11-16
Upgrade chart version to v5.0.3 for CVE-2019-11255

pure-yesmat:


operator-csi-plugin/README.md
```diff
@@ -113,6 +113,7 @@ The ``update.sh`` script is used to apply changes from ``values.yaml`` as follow
 ## Uninstall
 To uninstall the Pure CSI Operator, run 
 ```
+kubectl delete PSOPlugin/psoplugin-operator
```
sdodsley: Does this not happen anyway with the `delete all`?

operator-k8s-plugin/README.md
```diff
@@ -108,6 +108,7 @@ More details on using configuration labels can be found [here](../docs/flex-volu
 ## Uninstall
 To uninstall the Pure FlexVolume Operator, run
 ```
+kubectl delete PSOPlugin/psoplugin-operator
```
sdodsley: Same as above...

docs/index.yaml
```diff
@@ -18,7 +33,7 @@ entries:
     version: 1.0.2
   - apiVersion: v1
     appVersion: 1.0.1
```
sdodsley: Still 1.0.1 or should it be 1.0.3?

docs/index.yaml
```diff
@@ -18,7 +33,7 @@ entries:
     version: 1.0.2
```
sdodsley: As above

docs/index.yaml
```diff
@@ -1,9 +1,24 @@
...
+    - purestorage
+    maintainers:
+    - email: sales@purestorage.com
+      name: Pure Storage, Inc.
+    name: pure-csi
+    urls:
+    - https://purestorage.github.io/helm-charts/pure-csi-1.0.3.tgz
+    version: 1.0.3
   - apiVersion: v1
     appVersion: 1.0.2
```
sdodsley: Does this need to be here, or even changed to 1.0.3?

docs/index.yaml
```diff
@@ -18,7 +33,7 @@ entries:
     version: 1.0.2
   - apiVersion: v1
```
sdodsley: Maybe its the format in GitHub but there seem to be a lot of references to appversion and version

operator-csi-plugin/README.md
pure-yesmat: order matters, if you delete the children of the CRD it gets stuck deleting the CRD for some reason.

docs/index.yaml
pure-yesmat: I don't follow

operator-csi-plugin/README.md
pure-yesmat: although I think this is missing the namespace.

operator-csi-plugin/README.md
```diff
@@ -122,10 +123,6 @@ To completely remove the CustomResourceDefinition used by the Operator run
 ```
 kubectl delete crd psoplugins.purestorage.com
 ```
-If the CRD fails to delete you may be experiencing a known issue. Resolve this by running:
```
Pure-AdamuKaapan: Is this fixed by the extra `kubectl delete` at the top?

docs/index.yaml
Pure-AdamuKaapan: I think I see what @sdodsley is saying, it confused me too for a bit, but this looks correct. It looks like the helm chart has a list of all previous versions in it as well.

docs/index.yaml
```diff
@@ -18,7 +33,7 @@ entries:
...
     appVersion: 1.0.1
-    created: "2019-11-11T10:13:06.2762877-08:00"
+    created: 2019-11-15T15:56:15.0235638-08:00
```
Pure-AdamuKaapan: Did these change because you had to backport the fix or something similar?

docs/index.yaml
Pure-AdamuKaapan: I think this is part of the listing of past versions (see my comment response further down).

docs/index.yaml
dsupure: @Pure-AdamuKaapan is correct. Helm chart list all previous versions.

operator-csi-plugin/README.md
pure-yesmat: yep

docs/index.yaml
pure-yesmat: I actually have no idea how that changed!

pure-yesmat: This is automatically updated by helm. Looking at all our previous commits that seems to have always been the case!

pure-yesmat:
Would love to get some more eyes on this.

PR #143, opened 2019-12-10
Note in Readme that Fiber Channel is for bare-metal k8s

2vcps:


dsupure:
The changes look good. I wonder what will happen if k8s is running on a non-bare-metal environment?

PR #150, opened 2019-12-27
StorageClass FS Options support

dsupure:
**Summary of changes:**
For `pure` and `pure-block` storageclass.
1.  Add ` csi.storage.k8s.io/fstype` and `createoptions` in the parameter key-value map
2. Add `mountOptions`

No change for `pure-file`


dsupure:
> LGTM. Did you test it?

Thanks, the test results are fine.

Issue #142, opened 2019-12-05
These K8s charts are still using deprecated apiVersions


asfdtodd:
pso deployments are broken:

{"level":"error","ts":1575570508.7013617,"logger":"helm.controller","msg":"Release failed","namespace":"pso-operator","name":"psoplugin-operator","apiVersion":"purestorage.com/v1","kind":"PSOPlugin","release":"psoplugin-operator-5igmuk6pvf7od89wp3m0z75rr","error":"failed to install release: validation failed: [unable to recognize \"\": no matches for kind \"DaemonSet\" in version \"extensions/v1beta1\", unable to recognize \"\": no matches for kind \"Deployment\" in version \"apps/v1beta1\"]",

helm chart deployments are broken for the same reason.

It's been almost three months since the release of K8s 1.16 and I figured you guys would be on it by now.
https://kubernetes.io/blog/2019/09/18/kubernetes-1-16-release-announcement/

taherv:
Are you blocked by this ? If so, can I interest you to submit a patch ? :)

asfdtodd:
No, I'm not blocked. I forked it and updated the manifests myself and I'm using my own repo for now. I'm not familiar enough writing helm charts to attempt a patch and make a pull request. The maintainer should be able to knock this out no problem.

taherv:
Cool !

I would still encourage you to submit your patch, we can throw it at our internal CI system. Saves us having to redo some of the work you did.

Your choice.
We'll try to get to this sooner than later.

2vcps:
Also, the CSI Driver helm chart does not use the older API versions. Have you tried moving to CSI or is there something in the FlexDriver you need that is not in CSI yet?

asfdtodd:
> Also, the CSI Driver helm chart does not use the older API versions. Have you tried moving to CSI or is there something in the FlexDriver you need that is not in CSI yet?

According to the instructions there's no migration path from the flexdriver to CSI.

trevormeiss:
Fixed by https://github.com/purestorage/helm-charts/pull/151

PR #151, opened 2020-01-02
k8s Api updates for 1.16+

kengibous:
Several v1beta1 apis where deprecated in k8s 1.16+

This should help address issues like - https://github.com/purestorage/helm-charts/issues/142

Currently I am unable to use the plugin on my k8s 1.16.3 cluster

### DIFF
--- a/operator-csi-plugin/install.sh
+++ b/operator-csi-plugin/install.sh
@@ -99,7 +99,7 @@ fi
 counter=0
 TIMEOUT=10
 echo "
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: psoplugins.purestorage.com
@@ -139,7 +139,7 @@ fi
 # 3. Create RBAC for the PSO-Operator
 echo '
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: pso-operator
 rules:
@@ -295,7 +295,7 @@ rules:
 
 ---
 kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: pso-operator-role
 subjects:
@@ -309,7 +309,7 @@ roleRef:
 
 ---
 kind: Role
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: pso-operator
 rules:
@@ -362,7 +362,7 @@ rules:
 ---
 
 kind: RoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: default-account-pso-operator
 subjects:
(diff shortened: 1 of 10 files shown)


operator-csi-plugin/install.sh
```diff
@@ -99,7 +99,7 @@ fi
 counter=0
 TIMEOUT=10
 echo "
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
trevormeiss: Thank you for your contribution @kengibous!
Everything looks great except for the update to the CRD apiVersion, which would require users to upgrade their k8s cluster to 1.16. Are you ok reverting this change until more users upgrade?

kengibous:
Thanks for the feedback, I believe I can accomplish this using - https://helm.sh/docs/chart_template_guide/builtin_objects/ . I will work on this over the weekend.

kengibous:
@sdodsley added logic to helm charts and bash scripts that determine what api version of a resource to use

pure-k8s-plugin/templates/_helpers.tpl
```diff
@@ -17,3 +17,37 @@ release: {{ .Release.Name | quote }}
 {{ .Values.orchestrator.openshift.flexPath }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return the appropriate apiVersion for deployment.
+*/}}
+{{- define "deployment.apiVersion" -}}
+{{- if semverCompare ">=1.16-0" .Capabilities.KubeVersion.GitVersion -}}
```
dsupure: same change should be here.

pure-k8s-plugin/templates/_helpers.tpl
```diff
@@ -17,3 +17,37 @@ release: {{ .Release.Name | quote }}
...
+{{- else -}}
+{{- print "apps/v1beta1" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the appropriate apiVersion for daemonset.
+*/}}
+{{- define "daemonset.apiVersion" -}}
+{{- if semverCompare ">=1.16-0" .Capabilities.KubeVersion.GitVersion -}}
```
dsupure: and here

pure-csi/templates/_helpers.tpl
```diff
@@ -37,3 +37,47 @@ Create chart name and version as used by the chart label.
...
+{{- else -}}
+{{- print "apps/v1beta1" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the appropriate apiVersion for daemonset.
+*/}}
+{{- define "daemonset.apiVersion" -}}
+{{- if semverCompare ">=1.16-0" .Capabilities.KubeVersion.GitVersion -}}
```
dsupure: and here.

pure-k8s-plugin/templates/_helpers.tpl
kengibous: what change should here?

pure-k8s-plugin/templates/_helpers.tpl
kengibous: what change should here?

pure-k8s-plugin/templates/_helpers.tpl
dsupure: Hi, kengibous,
Sorry, I lost my first comment on it 😞 . Our current helm chart already uses `apps/v1` for daemonset. The change will make all versions below 1.16 falling back to use `apps/v1beta1`.
For deployment, we are still using 'apps/v1beta1'. However, it is ok to use `apps/v1`. 

I would recommend change `semverCompare ">=1.16-0` to `semverCompare ">=1.9-0` since 1.9 is the first version support v1. The change should have minimal impact to our customers. I've tested  `semverCompare ">=1.9-0` in our test infra and it passed all tests.

kengibous: thank you that helps a lot - thanks for mentioning that change, I will get that added

PR #152, opened 2020-01-03
Remove old beta reference in default storageclass definition

sdodsley:
Given we are using the v1 API for storage we should not be using the beta annotation name. This will be removed at some point in the future from the v1 API so we should use the correct annotation key.
The v1 API for storage was added back in v1.6 of k8s so we should have no issue with this change as this was prior to the minimum k8s version supported by PSO

### DIFF
--- a/pure-csi/templates/storageclass.yaml
+++ b/pure-csi/templates/storageclass.yaml
@@ -3,7 +3,7 @@ kind: StorageClass
 metadata:
   name: pure
   annotations:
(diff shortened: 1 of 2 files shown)


sdodsley:
@dsupure these are the only places we use it and the minimum version is in the commit message -v1.6

dsupure:
> @dsupure these are the only places we use it and the minimum version is in the commit message -v1.6

Nice! 1.6 is quite old. I just want to make sure it won't impact customers.

PR #157, opened 2020-01-17
CSI Release 1.0.5/5.0.5

dsupure:
**Summary of the release:**
* New CSI Feature: [Per-Volume FileSystem Options](https://github.com/purestorage/helm-charts/blob/master/docs/csi-filesystem-options.md)
* New CSI Feature: [Raw Block Volume Support](https://kubernetes-csi.github.io/docs/raw-block.html)
* Bug Fix: Add associated snapshots check when deleting a volume


operator-csi-plugin/values.yaml
```diff
@@ -4,7 +4,7 @@
 
 image:
   name: purestorage/k8s
-  tag: 5.0.4
```
Pure-AdamuKaapan: I wonder, is there a better way to set this version number everywhere? Maybe something to look into later...

dsupure: Yeah, it would be nice to have centralized control.  We can discuss later since the version number across multiple document types.

PR #159, opened 2020-01-17
Add option to override creation of built-in StorageClasses

bcdonadio:
This PR introduces the option for the chart user to not create the three built-in StorageClasses `pure`, `pure-file` and `pure-block`.

While this creation is handy for test clusters, production cluster will probably (like us) use a standardized StorageClass name based on the capabilities/purpose of each backing device. Not obligating the user to have these 3 classes when using the chart should help the sysadmins to maintain a clean list for the users.

### DIFF
--- a/pure-csi/templates/storageclass.yaml
+++ b/pure-csi/templates/storageclass.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.storageclass.createBuiltIn }}
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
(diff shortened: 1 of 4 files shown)


bcdonadio:
> Don't run the update script - just send the PR with the 4 files you actually changed

@sdodsley should be clean now.

dsupure:
@bcdonadio : This PR has been merged in our master branch via #203. I'll close this PR. Thanks for making our tool better.

PR #163, opened 2020-01-24
Support either string or list value for flasharray defaultMountOpt

trevormeiss:
When file system options were released in 5.0.5, the defaultMountOpt
in values.yaml changed from a string to a list.
For backwards compatibility, we should continue to support string values.

### DIFF
--- a/pure-csi/templates/storageclass.yaml
+++ b/pure-csi/templates/storageclass.yaml
@@ -12,12 +12,15 @@ parameters:
   backend: {{ .Values.storageclass.pureBackend | default "block" | quote }}
   csi.storage.k8s.io/fstype: {{ .Values.flasharray.defaultFSType | default "xfs" | quote }}
   createoptions: {{ .Values.flasharray.defaultFSOpt | default "-q" | quote }}
-
-{{- if .Values.flasharray.defaultMountOpt }}
+# support either string or list for .Values.flasharray.defaultMountOpt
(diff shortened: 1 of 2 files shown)


pure-csi/templates/storageclass.yaml
```diff
@@ -12,12 +12,15 @@ parameters:
...
-{{- if .Values.flasharray.defaultMountOpt }}
+# support either string or list for .Values.flasharray.defaultMountOpt
+{{- if or (kindIs "array" .Values.flasharray.defaultMountOpt) (kindIs "slice" .Values.flasharray.defaultMountOpt) }}
 mountOptions:
   {{- range .Values.flasharray.defaultMountOpt }}
   - {{ . }}
   {{- end }}
+{{- else if kindIs "string" .Values.flasharray.defaultMountOpt}}
+mountOptions:
+  - {{ .Values.flasharray.defaultMountOpt }}
```
dsupure: @trevormeiss: Can you try the string like this "discard stripe=32" and confirm it work? 
Thank you.

trevormeiss: The output for the env variable for the node server is:
```
           - name: PURE_DEFAULT_BLOCK_MNT_OPT
             value: "discard stripe=32"
```
The output for the storage class is:
```
parameters:
  backend: block
  csi.storage.k8s.io/fstype: xfs
  createoptions: -q
mountOptions:
  - discard stripe=32
```

PR #164, opened 2020-01-28
Added an orchestrator basePath value to the CSI chart

Pure-AdamuKaapan:
This is used to specify where the base Kubelet path is. The main use case is for something like CoreOS, where /var/lib/kubelet is read-only and a different directory must be used.

This has been tested on Kubernetes 1.14 and 1.16, both change the necessary parameters so that it works as expected. The Flex chart already has a similar value, so no changes are required.

### DIFF
--- a/pure-csi/templates/node.yaml
+++ b/pure-csi/templates/node.yaml
@@ -90,7 +90,7 @@ spec:
                 command: ["/bin/sh", "-c", "rm -rf /registration/pure-csi /registration/pure-csi-reg.sock"]
           args:
             - --csi-address=/csi/csi.sock
-            - --kubelet-registration-path=/var/lib/kubelet/plugins/pure-csi/csi.sock
+            - --kubelet-registration-path={{ .Values.orchestrator.basePath }}/plugins/pure-csi/csi.sock
           securityContext:
             privileged: true
           env:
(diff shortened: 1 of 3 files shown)


pure-csi/values.yaml
```diff
@@ -69,6 +69,9 @@ namespace:
 orchestrator:
   # name is either 'k8s' or 'openshift'
   name: k8s
+  # Use this to specify the base path of Kubelet. It should contain the "plugins", "plugins_registry", and "pods" directories.
+  # Default is /var/lib/kubelet. For CoreOS and Rancher, this is usually /opt/rke/var/lib/kubelet unless configured otherwise.
+  basePath: /var/lib/kubelet
```
caryli-ps: Per discussed offline, we probably need to gracefully handle the case which no value is provided by user's YAML. It could happen they just reuse an existing YAML file which was created a while back.

sdodsley: Agree. Sometimes customers have a `values.yaml` that only contains the parameters they care about, so we have to have defaults for all parameters in case they are not specified

Pure-AdamuKaapan: I'm still going to add a default value just for safety, but doesn't the helm chart also include this valuesl.yaml? So wouldn't the precedence be <user values.yaml> overrides <default values.yaml>?

dsupure: Helm chart should read value.yaml if it cannot find the value in the user's YAML. It is no harm to add the default to make it safer.

pure-csi/templates/node.yaml
```diff
@@ -204,19 +204,19 @@ spec:
             path: /dev
           name: dev
         - hostPath:
-            path: /var/lib/kubelet/plugins/pure-csi
+            path: {{ .Values.orchestrator.basePath }}/plugins/pure-csi
```
trevormeiss: don't we also want defaults for these values?

Pure-AdamuKaapan: Wow I'm dumb. Yes, thanks, good catch

PR #167, opened 2020-02-03
Conditionally install daemonset to configure nodes in GKE. 

neelakan79:
The daemonset invokes a script from the PSO image that is responsible for preparing GKE nodes for use. At the time of writing this includes installing open-iscsi and multipath-tools
as well as configuring and enabling iscsid.

Conditionally disable csi-snapshotter in GKE. A requirement for Anthos is that we do not enable any alpha features.

This PR depends on https://github.dev.purestorage.com/pso/k8s/pull/162

Together, this enables installation of PSO on Google Anthos and satisfies Google's tests for GKE 1.2.1.

### DIFF
--- a/pure-csi/templates/node-configure.yaml
+++ b/pure-csi/templates/node-configure.yaml
@@ -1,0 +1,25 @@
+# Automatic node configuration has only been tested on GKE.
+{{ if (.Capabilities.KubeVersion.GitVersion | regexMatch "gke") }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: pso-daemonset
+  namespace: {{ .Release.Namespace }}
+spec:
+  selector:
+    matchLabels:
+      name: pso-daemonset
+  template:
+    metadata:
+      labels:
+        name: pso-daemonset
+    spec:
+      hostPID: true
+      containers:
+      - name: pso-daemonset
+        image: "{{ .Values.image.name }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        command: ["/bin/sh", "-c", "/node-configure.sh"]
+        securityContext:
+          privileged: true
+{{ end }}
--- a/pure-csi/templates/provisioner.yaml
+++ b/pure-csi/templates/provisioner.yaml
@@ -78,6 +78,8 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+# Google Anthos (which is built on GKE) prohibits alpha snapshotters
+{{- if not (.Capabilities.KubeVersion.GitVersion | regexMatch "gke") }}
         - name: csi-snapshotter
           {{- with .Values.csi.snapshotter.image }}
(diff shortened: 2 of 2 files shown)


pure-csi/templates/node-configure.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      hostPID: true
+      containers:
+      - name: pso-daemonset
+        image: "{{ .Values.image.name }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        command: ["/bin/sh", "-c", "/node-configure.sh"]
+        securityContext:
+          privileged: true
+{{ end }}
```
sdodsley: Missing EOL

pure-csi/templates/provisioner.yaml
```diff
@@ -78,6 +78,7 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+{{- if not (.Capabilities.KubeVersion.GitVersion | regexMatch "gke") }}
```
sdodsley: Why are we not using the snapshotter in GKE? 
If it is to do with lack of support, then I'm sure GKE will support it at some point, so we need a version check, not just a blanket no to GKE

taherv: Right now it is kind of like a blanket "no". 
Reasons:
a. The snapshotter sidecar used by PSO is an alpha version.
b. Anthos requirement is no alpha snapshotters. Apparently, there is a serious upgrade issue of the alpha snapshotter in k8s v1.17 and Anthos wants to avoid that problem.
c. Currently released versions of Anthos have no snapshotter support (even though CSI drivers can claim they have snapshot support).

neelakan79: At the moment Anthos explicitly requires us to disable the csi-snapshotter. We fail certification without this check.

If and when Anthos supports csi-snapshotter we can add a version check.

taherv: > then I'm sure GKE will support it at some point,
We should totally have a better `if` condition with a version check when that happens.

trevormeiss: I guess since we will also have to update the sidecar version I'm ok with waiting on a better `if` condition. However I still feel that at least a comment would be great.

pure-csi/templates/node-configure.yaml
neelakan79: Fixed

pure-csi/templates/provisioner.yaml
neelakan79: Added a couple comments

neelakan79:
I resolved the conversations that were in flight, please holler if I didn't resolve your concerns.

taherv:
Wow ! 3 approvers ... thanks a lot !
Can we merge this now ?

taherv:
Since this PR is approved, and its dependent PR (https://github.dev.purestorage.com/pso/k8s/pull/162) is also merged in .... going ahead with this merge.

PR #169, opened 2020-02-04
Update README.

taherv:
Document support for Google Anthos


taherv:
Some context on this PR.
In an upcoming announcement in Feb, Google will announce Pure as a Anthos partner. In that announcement, they want to link to some documentation provided by Pure that guides customers on how to install Pure's CSI driver.
I wanted to add this patch to update our install instructions, saying Anthos is now a supported platform by this helm chart.

Since the code to support Anthos has gone in, and our manual tests for Anthos are passing, this documentation can be safely updated. There's not much to documentation for Anthos, beyond just saying that it is supported, which is all that this patch is about.

Hope this helps reviewers.....

PR #181, opened 2020-02-27
Should the snapshotclass.yaml be moved?

2vcps:
The snapshotclass.yaml is not installed by default. If there is a reason for this, that is fine. I was using an application that needs access to the snapshotclass and was surprised this is a manual step. Moving the yaml to the templates folder should allow it to install automatically, if that is desired.


taherv:
I don't think it should be moved to be installed automatically. Reasons: 
a. It can void the warranty on some production k8s clusters, since the volumesnapshotclass CRD is alpha for some versions of k8s. 
b. Dont know the specifics but the alpha snapshot CRDs also have an upgrade problem since k8s 1.17. 
c. Also, it will break Anthos automated tests.

2vcps:
Thanks @taherv that makes sense now.

sdodsley:
I do think you closed this a little early.
There could have been some discussion on this. 
I am sure there are ways we could mitigate.
At the very least we need to write some documentation on the process to follow, although I'm sure we had this enabled by default at one point and I don't remember seeing it being disabled...

dsupure:
You need the `volumesnapshotclasses CRD` before you can install the `volumesnapshotclass`. The `volumesnapshotclasses CRD` is installed by the external snapshotter sidecar. The snapshot sidecar is in our csi-provisioner pod. This means you can only install the `volumesnapshotclass` after the csi-provisioner is installed. I tried to put the `snapshotclass.yaml` in the templates and it complained it could not find the `volumesnapshotclasses CRD`.  Helm doesn't provide the control of the order of installation. So we discussed to install it manually if you use helm. However, it will install it automatically if you use the CSI operator: 

https://github.com/purestorage/helm-charts/blob/master/operator-csi-plugin/install.sh#L466-L482

PR #182, opened 2020-02-27
Fix v1beta1 API version for the PSOPlugin CRD

dsupure:
The current operator installation breaks our pso driver.  This is because the `install.sh` script fails to generate the correct PSOPlugin object when using `apiextensions.k8s.io/v1` CRD API to create the `PSOPlugin CRD`. This fix rolls back to use ` apiextensions.k8s.io/v1beta1`.  

I've tested this PR using CI's functional tests on k8s 1.15 and 1.17 version. Both are tested fine.
Note that `apiextensions.k8s.io/v1beta1` is going to depreciate in 1.19 release.  For the future 1.19 release, we need to add `openAPIV3Schema` for strong typing validation of our parameters in the `values.yaml` file.

### DIFF
--- a/operator-csi-plugin/install.sh
+++ b/operator-csi-plugin/install.sh
@@ -105,34 +105,8 @@ fi
 # 2. Create CRD and wait until TIMEOUT seconds for the CRD to be established.
 counter=0
 TIMEOUT=10
-
-if [[ ${CRDAPIVERSION} == "apiextensions.k8s.io/v1" ]]; then
-    echo "
-apiVersion: ${CRDAPIVERSION}
-kind: CustomResourceDefinition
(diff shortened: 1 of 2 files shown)


dsupure:
> LGTM, could you test on k8s 1.15 and 1.16 just to be 100% sure?

k8s1.15 has been tested. I'll test 1.16.

dsupure:
> I assume you meant that the beta API is deprecating in 1.19 in the PR comment...

Yeah. Update the Conversation. Sorry for the confusion.

dsupure:
k8s 1.16 test passed. I'll merge it now and ask Jiafeng to do the release test.

PR #184, opened 2020-02-28
Update Documentation - Helm2/3 changes and formatting

sdodsley:
Some **MAJOR** support statement changes here...

* Only OpenShift 3.11 is supported given Red Hat finished support for 3.10 and older in Oct 2019.
* Only Operator install for FlexDriver is supported in OpenShift 3.11, given Helm2 has moved into maintenance mode and will be deprecated later in the year,
* FlexDriver install is not supported with Helm3
* Add requirement to create a namespace for PSO before running Helm(3) install
* Remove the `labels:` section from the `arrays:` section of `values.yaml` for the CSI driver. These can be added back in when they are supported by Storage Topology - assuming we do it the same way...

also, some formatting fixes for markdown files

### DIFF
--- a/pure-csi/README.md
+++ b/pure-csi/README.md
@@ -11,7 +11,7 @@ This helm chart installs the CSI plugin on a Kubernetes cluster.
   - Ubuntu 18.04
 - #### Environments Supported*:
   - Kubernetes 1.13+
-  - Minimum Helm version required is 3.0.3.
+  - Minimum Helm version required is 3.1.0.
   - OpenShift 3.11
   - Google Anthos 1.2.x
 - #### Other software dependencies:
@@ -115,23 +115,15 @@ arrays:
   FlashArrays:
     - MgmtEndPoint: "1.2.3.4"
       APIToken: "a526a4c6-18b0-a8c9-1afa-3499293574bb"
-      Labels:
-        rack: "22"
-        env: "prod"
     - MgmtEndPoint: "1.2.3.5"
       APIToken: "b526a4c6-18b0-a8c9-1afa-3499293574bb"
   FlashBlades:
     - MgmtEndPoint: "1.2.3.6"
       APIToken: "T-c4925090-c9bf-4033-8537-d24ee5669135"
       NfsEndPoint: "1.2.3.7"
-      Labels:
-        rack: "7b"
-        env: "dev"
     - MgmtEndPoint: "1.2.3.8"
       APIToken: "T-d4925090-c9bf-4033-8537-d24ee5669135"
       NfsEndPoint: "1.2.3.9"
-      Labels:
-        rack: "6a"
 ```
 
 ## Assigning Pods to Nodes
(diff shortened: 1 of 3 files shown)


pure-csi/README.md
```diff
@@ -115,23 +115,15 @@ arrays:
   FlashArrays:
     - MgmtEndPoint: "1.2.3.4"
       APIToken: "a526a4c6-18b0-a8c9-1afa-3499293574bb"
-      Labels:
```
Pure-AdamuKaapan: I assume we're removing this right now because CSI doesn't support importing/anything with labels at the moment?

pure-k8s-plugin/README.md
```diff
@@ -9,11 +9,10 @@ This helm chart installs the FlexVolume plugin on a Kubernetes cluster.
   - CoreOS (Ladybug 1298.6.0 and above)
   - Ubuntu 16.04
   - Ubuntu 18.04
-  - For Platform specific requirements see
 - #### Environments Supported*:
   - Kubernetes 1.6+
-  - Minimum Helm version required is 3.0.3
-  - [OpenShift](#openshift) 3.6+
+  - Helm 2.9.1+ (**NOTE:** Helm3 is not supported for FlexDriver)
```
Pure-AdamuKaapan: Why is this?

pure-k8s-plugin/README.md
```diff
@@ -257,13 +262,15 @@ Once changed the kublet parameters need to be updated to set the
 via the `flexPath` option in your `values.yaml`.
 
 ### CoreOS
+
 Similar to the Atomic hosts this has a read-only `/usr` tree and requires
```
Pure-AdamuKaapan: Note that one of our earlier PRs has the "kubeletBasePath" change, which is relevant to CoreOS (and I assume Atomic). Dunno if we want to add that in or just tell them to keep using the existing instructions.

pure-csi/README.md
sdodsley: Yes. As it says in the PR description...

pure-k8s-plugin/README.md
sdodsley: Read the description for the PR

pure-k8s-plugin/README.md
sdodsley: I am close to removing this CoreOS section anyway given it is deprecated in favour of Red Hat/Fedora CoreOS, which we will probably have to validate at some point

pure-csi/README.md
Pure-AdamuKaapan: Ope, I'm so sorry, I dunno how I missed the description. That's my bad.

pure-csi/README.md
```diff
@@ -145,6 +137,12 @@ For security reason, it's strongly recommended to install the plugin in a separa
 
 Customize your values.yaml including arrays info (replacement for pure.json), and then install with your values.yaml.
 
+Create a namespace for PSO to install into
+
+```bash
+kubectl createnamespace <namespace>
```
caryli-ps: I think it's missing a space between `create namespace`

PR #185, opened 2020-03-04
Release 5.0.8/2.5.8

dsupure:
**Summary of changes for the Release 5.0.8/2.5.8 **

1. Rollback APIVersion from v1 to v1beta1 for PSOplugin for the PSO csi/flex operators. V1 needs a strong typing schema and we need time to bake it. 
2. Fix RoleBinding API version to "rbac.authorization.k8s.io/v1beta1" for openshift.


operator-csi-plugin/install.sh
```diff
@@ -52,12 +52,16 @@ while (("$#")); do
         ;;
     esac
 done
-
-CRDAPIVERSION="$(${KUBECTL} explain CustomResourceDefinition | grep "VERSION:" | awk '{ print $2 }')"
+# Disable CRDAPIVERSION because openshift 3.11 does not support it.
```
sdodsley: Just because OpenShift 3.11 doesn't support CRDs doesn't mean that other k8s deployments, including OpenShift 4.x don't either.
You can't just comment this out.

operator-k8s-plugin/install.sh
```diff
@@ -52,12 +52,16 @@ while (("$#")); do
         ;;
     esac
 done
-
-CRDAPIVERSION="$(${KUBECTL} explain CustomResourceDefinition | grep "VERSION:" | awk '{ print $2 }')"
+# Disable CRDAPIVERSION because openshift 3.11 does not support it.
```
sdodsley: Same as above

operator-csi-plugin/install.sh
dsupure: We don't use this `CRDAPIVERSION` variable even in the k8s deployments today.
The previous PR already take it out: https://github.com/purestorage/helm-charts/commit/ac469f57f7b043dce8e824975a0847ab58da060d. The reason I comment this out is we are working on a new internal CI/CD release test pipeline and 
``` bash
oc explain CustomResourceDefinition
``` 
will return an error code. Let me remove it since we don't use it anywhere.

sdodsley: OK

PR #186, opened 2020-03-07
Update snapshot docs

sdodsley:
Add clarification over volume snapshot recovery

### DIFF
--- a/docs/csi-snapshot-clones.md
+++ b/docs/csi-snapshot-clones.md
@@ -154,7 +154,7 @@ spec:
     apiGroup: snapshot.storage.k8s.io
 ```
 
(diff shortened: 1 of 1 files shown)


docs/csi-snapshot-clones.md
```diff
@@ -154,6 +154,12 @@ spec:
     apiGroup: snapshot.storage.k8s.io
 ```
 
+**NOTE:** Recovery of a volume snapshot to overwite its parent persistant volume is not supported in he CSI specification, however this can be achieved with a FlashArray based PVC and snapshot using the following steps:
```
dsupure: Should be `is not supported in the`

PR #187, opened 2020-03-12
Add CSI Topology Support

dsupure:
**Summary of changes:**
* Add CSI Topology Support document.

### DIFF
--- a/docs/csi-topology.md
+++ b/docs/csi-topology.md
@@ -13,22 +13,39 @@ It also allows topology to be further specified or constrained for both pre-prov
 The following dependencies must be true before the storage topology functionality can be used:
 
 * Kubernetes already running, deployed, configured, etc
-* PSO correctly installed and using [Pure CSI Driver v6.0.0](https://github.com/purestorage/helm-charts/releases/tag/6.0.0)+.
+* PSO correctly installed and using [Pure CSI Driver v5.0.9](https://github.com/purestorage/helm-charts/releases/tag/5.0.9)+.
 * Ensure you have Kubernetes 1.13+ installed, if you are using a version less than 1.13, please make sure the `VolumeScheduling` feature gate is enabled
 
 ## Caveats
-* Please make sure your backend arrays supports all topologies defined in your system. Failing to do so will potential leak volumes. 
-PSO follows the [CSI Spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) to do its best effort to provision a volume with all possible preferred topologies specified by the scheduler.
-For example, if you wants to provision a pod to be in `region-2` by assigning the topology via `nodeAffinity`, scheduler will ask PSO to provision a volume with the preferred topologies like [`region-2` , `region-1`, `region-0`]. In this case, if your arrays don't support `region-2`, PSO end up to provision the volume in the one of the remaining regions [`region-1`, `region-0`].
-The scheduler will not provision the pod because the the volume's topology does not match the pod's requirement. This is an expected behavior, but it leaks the volume and you need to manually delete the created volume.
+* Please make sure your backend arrays support all topologies defined in your system. Failing to do so will potentially leak volumes.
+  PSO follows the [CSI Spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) to do its best effort to provision a volume with all possible topologies specified by the scheduler.
+For example, if you want to provision a pod to be in `region-2` by assigning the topology via `nodeAffinity`, the scheduler will ask PSO to provision a volume with the topology order like this [`region-2` , `region-1`, `region-0`]. In this case, if your arrays don't support `region-2`, PSO ends up to provision the volume in one of the remaining regions [`region-1`, `region-0`].
+The scheduler will end up failing to provision the pod because the volume's topology does not match the pod's requirement. Although this is expected behavior, it leaks the unexpected volume, and you need to delete it manually.
+
 ## How to Use The Topology-Aware Provisioning Feature in PSO
-It is straightforward to use the topology feature in PSO. PSO provides five pre-defined topology keys to help you label your backend arrays, such as FlashArray for FlashBlade, as well as pods and nodes resources in Kubernetes.
-These topology keys are:
-* `topology.purestorage.com/region`
-* `topology.purestorage.com/zone`
-* `topology.purestorage.com/rack`
-* `topology.purestorage.com/name`
-* `topology.purestorage.com/env`
+There are two parameters in the `values.yaml` to enable/disable the CSI topology :
+```yaml
+storagetopology:
+  enable: false
+  strictTopology: false
+```
+`enable` controls enable/disable the CSI topology and `storagetopology` sets  the [strict-topology feature](https://github.com/kubernetes-csi/external-provisioner/blob/master/README.md#topology-support) when users need it. 
+ They are set to `false` by default. You need to set the parameters to `true` to enable the feature.
+
+It is straightforward to use the topology feature in PSO. PSO provides users a pre-defined **pure topology key prefix** (`topology.purestorage.com/`) to label any topology segments you want.
+You can define your topology key/value pairs like these:
+
+```yaml
+topology.purestorage.com/foo : "foo"
+topology.purestorage.com/bar : "bar"
+
+```
+or more formal cloud-hierarchical ways: 
+```yaml
+topology.purestorage.com/zone : "zone-0"
+topology.purestorage.com/region : "region-0"
+```
+
 
 ### Example of Adding Topology Labels to Arrays
 To add topology labels to your backend array, please add them in the `Labels` section of you `values.yaml` file.
@@ -67,9 +84,8 @@ kubectl label node k8s-cluster-0 topology.purestorage.com/zone=zone-1
 kubectl label node k8s-cluster-0 topology.purestorage.com/region=region-1
 kubectl label node k8s-cluster-0 topology.purestorage.com/env=dev
 ``` 
-**NOTE:** Please make sure you label the cluster node topology before installing the PSO. During the PSO initialization phase, PSO will talk to Kubernetes to retrieve topology labels from nodes and generate a `CSINode` object for future reference.
-The topology information in the `CSINode` will not update at runtime, which means any node topology label update will not affect the initialized values.
-      
+**NOTE:** PSO supports dynamic topology labeling. It allows users to add/remove topology labels of cluster nodes on the fly without reloading the PSO. However, users still need to update the `values.yaml` and reload the driver if you want to change the backend array topologies.
+
 ## Delayed Volume Binding
 Without the topology-aware feature, volume binding occurs immediately once a PersistentVolumeClaim is created. For volume binding to take into account all of a pod’s other scheduling constraints, volume binding must be delayed until a Pod is being scheduled.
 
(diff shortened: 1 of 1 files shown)


docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
...
+
+The Pure Service Orchestrator Kubernetes CSI driver includes support for topology. This feature allows Kubernetes to make intelligent decisions when dynamically provisioning volumes by getting scheduler input on the best place to provision a volume for a pod.
+For example, in multi-zone clusters, volumes will get provisioned in an appropriate zone that can run your pod, allowing you to quickly deploy and scale your stateful workloads to provide high availability and better fault tolerance.
+
+It also allows topology to be further specified or constrained for both pre-provisioned and dynamic provisioned PersistentVolumes (PV)  so that the Kubernetes scheduler can correctly place a Pod using such a volume to an appropriate node.
+  
+
+## Dependencies
+
+The following dependencies must be true before the snapshot and clone functionality can be used:
```
caryli-ps: [Question]
Do we need this line with "snapshot and clone functionality" here?

docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
...
+
+* Kubernetes already running, deployed, configured, etc
+* PSO correctly installed and using [Pure CSI Driver v6.0.0](https://github.com/purestorage/helm-charts/releases/tag/6.0.0)+.
+* Ensure you have Kubernetes 1.13+ installed, if you are using a version less than 1.13, please make sure the `VolumeScheduling` feature gate is enabled
+
+## Caveats
+* Please make sure your backend arrays supports all topologies defined in you system. Failing to do so will potential leak volumes, since
```
caryli-ps: [minor]
spelling - "in your system"

docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
...
+* `topology.purestorage.com/region`
+* `topology.purestorage.com/zone`
+* `topology.purestorage.com/rack`
+* `topology.purestorage.com/name`
+* `topology.purestorage.com/env`
+
+### Example of Adding Topology Labels to Arrays
+To add topology labels to your backend array, please add them in the `Labels` section of you `values.yaml` file.
+
+```yaml
```
caryli-ps: [Question]
YAML or JSON?

I think we want to attach the `values.yaml` section of this.

docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
...
+``` 
+
+### Example of Labeling Topology to Cluster Nodes
+
+```bash
+kubectl label node k8s-cluster-0 topology.purestorage.com/zone=zone-1
+kubectl label node k8s-cluster-0 topology.purestorage.com/region=region-1
+kubectl label node k8s-cluster-0 topology.purestorage.com/env=dev
+``` 
+**NOTE:** Please make sure you label the cluster node topology before installing the PSO. During the PSO initialization phase, PSO will talk to Kubernetes to retrieve topology labels from nodes and generate a `CSINode` object for future reference.
```
caryli-ps: [Major]
This seems like a major drawback. 

Let's say if customer really updated cluster node topology, is there anything they can do to restart PSO to refresh topology labels? Maybe by resizing replica count to 0, then resize back to 3?

sdodsley: I tend to agree with this comment. What happens when a new node is added to the cluster. The daemonset may well install PSO before the node topology can be set by the cluster admin as well.
We need a way to tell CSI to rescan the topology requirements. 
With new arrays added it was just a case of running an upgrade script we provided. Could this not do the same for node topology checking?

docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
+# Using CSI Topology-Aware Volume Provisioning with Kubernetes
+
+## Introduction
+
+The Pure Service Orchestrator Kubernetes CSI driver includes support for topology. This feature allows Kubernetes to make intelligent decisions when dynamically provisioning volumes by getting scheduler input on the best place to provision a volume for a pod.
```
sdodsley: storage topology

docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
...
+For example, in multi-zone clusters, volumes will get provisioned in an appropriate zone that can run your pod, allowing you to quickly deploy and scale your stateful workloads to provide high availability and better fault tolerance.
+
+It also allows topology to be further specified or constrained for both pre-provisioned and dynamic provisioned PersistentVolumes (PV)  so that the Kubernetes scheduler can correctly place a Pod using such a volume to an appropriate node.
```
sdodsley: 1. dynamically, not dynamic
2. remove extra whitespace after `(PV)`

docs/csi-topology.md
sdodsley: Shouldn't this read `storage topology functionality`?

docs/csi-topology.md
sdodsley: May want to clarify this whole sentence, especially the reference to leak volumes

docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
...
+* PSO correctly installed and using [Pure CSI Driver v6.0.0](https://github.com/purestorage/helm-charts/releases/tag/6.0.0)+.
+* Ensure you have Kubernetes 1.13+ installed, if you are using a version less than 1.13, please make sure the `VolumeScheduling` feature gate is enabled
+
+## Caveats
+* Please make sure your backend arrays supports all topologies defined in you system. Failing to do so will potential leak volumes, since 
+PSO follows the [CSI Spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) to do its best effort to provision a volume for you using the remaining preferred topologies specified by the scheduler.
+
+## How to Use The Topology-Aware Provisioning Feature in PSO
+It is straightforward to use the topology feature in PSO. PSO provides five pre-defined topology keys to help you label your backend arrays, such as FlashArray for FlashBlade, as well as pods and nodes resources in Kubernetes.
+These topology keys are:
```
sdodsley: Are we limited to these five? What happens is the customer wants more granular control and needs to create more arbitrary keys? In the Flex version of this, we could link to any key.

docs/csi-topology.md
sdodsley: Given this is a README type document, the example should be should exactly as the `values.yaml` file would be seen. We don't have all these parentheses in that file...

docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
...
+The topology information in the `CSINode` will not update at runtime, which means any node topology label update will not affect the initialized values.
+      
+## Delayed Volume Binding
+Without the topology-aware feature, volume binding occurs immediately once a PersistentVolumeClaim is created. For volume binding to take into account all of a pod’s other scheduling constraints, volume binding must be delayed until a Pod is being scheduled.
+
+A new StorageClass field `volumeBindingMode` is introduced to control the volume binding behavior.
+You can specify two values:
+
+* `Immediate`: This the default binding method. External-provisioner will pass in all available topologies in the cluster for the driver.
+* `WaitForFirstConsumer`: external-provisioner will wait for the scheduler to pick a node. The topology of that selected node will then be set as the first entry in  CreateVolumeRequest.accessibility_requirements.preferred . All remaining topologies are still included in the requisite and preferred fields to support storage systems that span across multiple topologies.
```
sdodsley: check format of CreateVolumeRequest.accessibility_requirements.preferred. Should be `CreateVolumeRequest.accessibility_requirements.preferred`

docs/csi-topology.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Example of StorageClass with Delayed Volume Binding 
+```yaml
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: pure-block-delay-binding
+  labels:
+    kubernetes.io/cluster-service: "true"
+    generator: helm
```
sdodsley: This should not be required

docs/csi-topology.md
dsupure: Yeah, my bad. It should be `storage topology functionality`

docs/csi-topology.md
dsupure: I don't have the reference, but I add an example to explain the leak volumes. Do we want to call it `leak volumes` or we have a better term to describe this scenario.

docs/csi-topology.md
dsupure: hmm.. yeah this is JSON format. Let me switch to yaml.

docs/csi-topology.md
dsupure: The topology information is stored in the `CSINode` object, which is generated from the `NodeGetInfo` CSI API. When a new node is added in the system, kubelet will call `NodeGetInfo` to get the topology information. The cloud vendors have their "NodeMetaService". When the `NodeGetInfo` API is called, it will query the "NodeMetaService" to get the topology information of that node so they can scale up/down to force CSI to rescan topology.   

Since we don't have "NodeMetaService" in our system, I come up with this labeling solution, but this is a major drawback because the daemonset may install PSO before the node topology can be set as Simon said. 

I am thinking of another workaround and it should work:
When calling `NodeGetInfo`, we return the `nodeID` instead of topologies in the `AccessibleTopology` field. The CSI passes `AccessibleTopology` field to our driver when provisioning a volume, we can use the nodeID to talk to the API server to get its topology information during each volume provision call. That we can always get the latest topology of nodes.

docs/csi-topology.md
dsupure: Fixed.

docs/csi-topology.md
dsupure: What flex provides is the label selector, not a topology. A topology consists of multiple segments such as region, zone, rack, that organize the segments in a hierarchical way. I checked most of the CSI topology implementations use the pre-defined keys instead of using the label selector. We can add the label selector if we need the needs from customers, however it against the origin SCI topology design concept.

docs/csi-topology.md
sdodsley: when I know what you mean by a `leak volume` I can better determine if we need a different name

docs/csi-topology.md
sdodsley: You could say that any selector is a topological view. Just because the CSI team didn't come up with it doesn't mean its not valid.
Can each topology be a list, or is it limited to a single item?
Could `topology.purestorage.com/env` be 'uat, dev, preprod`??

dsupure: Right now it is limited to a single item.
I think we can add a new key like this `topology.purestorage.com/label_selector` to support a list.
```bash
topology.purestorage.com/label_selector=uat,dev,preprod
```

sdodsley: OK - let's try that and check it works. An example of the PSO config side and the host kubectl command would be helpful as well (assuming it all works)

docs/csi-topology.md
dsupure: code change for the workaround is done. testing it now.

docs/csi-topology.md
dsupure: Forget about `topology.purestorage.com/label_selector=uat,dev,preprod`.

I rewrote the codes and we can label any key/value for the topology now. We don't need any pre-defined key anymore. The only requirement is the key need to have the prefix `topology.purestorage.com/`

You can assign the topology labels like 
```
topology.purestorage.com/foo=foo-0
topology.purestorage.com/boo=boo-0
```

docs/csi-topology.md
dsupure: I called this feature "dynamic labeling" I have an end-to-end functional test and it is working fine. I'll add a section to describe this feature.

docs/csi-topology.md
dsupure: The feature is done. We can label any key/value pair as long as you have `topology.purestorage.com/` prefix in the key.

sdodsley:
Need to add something about the fact you can use any key and they can be a list...

PR #188, opened 2020-03-13
Add --node-selector flag to operator-k8s-plugin/install.sh for openshift.

pure-jliao:
for CI testing purpose, the `--node-selector` should be empty in the CI test

### DIFF
--- a/operator-k8s-plugin/README.md
+++ b/operator-k8s-plugin/README.md
@@ -56,11 +56,9 @@ The install script will do the following:
 1. Create New Project.<br/>
 The script creates a new project (if it does not already exist) with the given namespace. If no namespace parameter is specified, the ``pso-operator`` namespace is used.<br/> 
 **OpenShift Note**: In OpenShift 3.11, the default node-selector for a project does not allow PSO Operator to mount volumes on master and infra nodes. 
-If you want to mount volumes on master and infra nodes OR run pods in the default namespace using volumes mounted by PSO, then modify the install script as follows.<br/>
+If you want to mount volumes on master and infra nodes OR run pods in the default namespace using volumes mounted by PSO, then set `--node-selector` flag to `""` when running the install script as follows.<br/>
 ```
-Existing line:     $KUBECTL adm new-project ${NAMESPACE}
-
-Change to allow volume mounts on master and infra nodes:     $KUBECTL adm new-project ${NAMESPACE} --node-selector=""
+install.sh --image=<image> --namespace=<namespace> --orchestrator=<ochestrator> --node-selector=<node-selector> -f <values.yaml>
 ```
 
 2. Create a Custom Resource Definition (CRD) for the PSO Operator. <br/>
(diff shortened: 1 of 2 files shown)


pure-jliao:
@sdodsley - `--node-selector==""` is required for our internal functional tests, but customer may not want it due to security reason. So what should be the most common case for this? Does this change make sense, or we should use the default `node-selector` by default, only set it to empty in our CI test?

operator-k8s-plugin/install.sh
```diff
@@ -3,10 +3,11 @@ IMAGE=quay.io/purestorage/pso-operator:v0.0.11
 NAMESPACE=pso-operator
 KUBECTL=oc
 ORCHESTRATOR=k8s
+NODESELECTOR=""
 
 usage()
 {
-    echo "Usage : $0 --image=<imagename> --namespace=<namespace> --orchestrator=<orchestrator> -f <values.yaml>"
+    echo "Usage : $0 --image=<imagename> --namespace=<namespace> --orchestrator=<orchestrator> --node-selector=<node-selector> -f <values.yaml>"
```
dsupure: node-selector is only used in openshift. This message will confuse k8s users.

pure-jliao: yes, simply removed this.

pure-jliao:
> We should not mess with the default `node-selector` setting. This will go against the prefered OpenShift standard.
> We should keep the current docs comment about how to allow PSO to run on master and infra nodes however.
> I would say only set this to be blank for our internal testing CI suite.

Make sense, updated accordingly.

operator-k8s-plugin/install.sh
```diff
@@ -3,6 +3,7 @@ IMAGE=quay.io/purestorage/pso-operator:v0.0.11
 NAMESPACE=pso-operator
 KUBECTL=oc
 ORCHESTRATOR=k8s
+NODESELECTOR="default"
```
dsupure: Is this the default value used in the openshfit? Why don't we use a more unique string like "unset" ?

pure-jliao: not used, just an indicator for whether we should use an external parameter. yep `unset` might be better.

operator-k8s-plugin/README.md
```diff
@@ -56,11 +56,9 @@ The install script will do the following:
 1. Create New Project.<br/>
 The script creates a new project (if it does not already exist) with the given namespace. If no namespace parameter is specified, the ``pso-operator`` namespace is used.<br/> 
 **OpenShift Note**: In OpenShift 3.11, the default node-selector for a project does not allow PSO Operator to mount volumes on master and infra nodes. 
-If you want to mount volumes on master and infra nodes OR run pods in the default namespace using volumes mounted by PSO, then modify the install script as follows.<br/>
+If you want to mount volumes on master and infra nodes OR run pods in the default namespace using volumes mounted by PSO, then set `--node-selector` flag to `""` when run the install script as follows.<br/>
```
sdodsley: running, not run

pure-jliao: changed.

PR #194, opened 2020-03-22
Add csi storage topology support

dsupure:
**Summary of changes:**
Add CSI storage topology support

### DIFF
--- a/pure-csi/templates/provisioner.yaml
+++ b/pure-csi/templates/provisioner.yaml
@@ -75,6 +75,9 @@ spec:
           args:
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
+{{- if eq .Values.storagetopology.enable true}}
+            - --feature-gates=Topology=true
+{{- end}}
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
--- a/pure-csi/templates/provisioner.yaml
+++ b/pure-csi/templates/provisioner.yaml
@@ -75,8 +75,9 @@ spec:
           args:
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
-{{- if eq .Values.storagetopology.enable true}}
-            - --feature-gates=Topology=true
+            - --feature-gates=Topology={{ .Values.storagetopology.enable }}
+{{- if and .Values.storagetopology.strictTopology .Values.storagetopology.enable}}
+            - --strict-topology
 {{- end}}
           volumeMounts:
             - name: socket-dir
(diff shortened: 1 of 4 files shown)


pure-csi/templates/provisioner.yaml
```diff
@@ -75,6 +75,9 @@ spec:
           args:
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
+{{- if eq .Values.storagetopology.enable true}}
```
sdodsley: What happens when this is no longer feature gated? 
With this we have to remember to keep a track of it and change when topology becomes GA, but then we will also have to add a k8s version check for lower than GA... 
Can we add a version check instead of a parameter? I know we don't know the GA version yet, but we can put a placeholder in for the time being...

dsupure: This feature gate is for the external provisioner, not for the k8s. Since we use external provisioner 1.4, it already supports the topology.  The minimal version of k8s to support topology is 1.14. When the version is less k8s 1.14, the external provisioner will ignore the topology feature even you enabled it. So we don't need to check the k8s version here.

dsupure: Also, I think we **SHOULD** set this value `false` by default no mater this feature gate is GA or not.

If we set to `true` by default, it will change the provisioning behavior when we try to mount more than one volumes in the same pod and we don't want it: 
 
Assuming we have three nodes clusters, each node is in `zone-0`, `zone-1`, `zone-2`, and we want to provision a pod with two volumes. In this case, the scheduler will send the provision request with the topology order like this:
```
volumes, topology order[ zone-0, zone-1, zone-2]
volume1, topology order [ zone-1, zone-2, zone-0]
 ```
In this case, you will see the pod is in the pending state with an error message and complains volume node affinity conflict.

There are two solutions for it:
1. Set --[strict-topology](https://github.com/kubernetes-csi/external-provisioner/blob/master/README.md#topology-support). Passes only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.

2. Set node affinity.

The topology feature needs more mature k8s and storage admins to use it and set it `false` by default can avoid lots of issues.

sdodsley: OK - If the switch is ignored by the ext. prov. then I'm fine with the check you have.
I"m also OK with the default being `false`. We may need to update the doc page for this to clarify about `allowedTopologies` and the use of `strict-topology` .
Other than this - LGTM

dsupure: Sounds good, I'll update it accordingly.

PR #196, opened 2020-03-26
Add CSI storage topology support

dsupure:
**Summary of change:**
Add CSI storage topology support. 

My previous change did not work when using the operator to install due to the pso-operator RBAC permission. This change has been tested using helm-install and operator install methods in our new test pipeline. The new pipeline can capture the regression in the helm chart which we could not do it before. The release quality will be much better moving forward.

### DIFF
--- a/docs/csi-topology.md
+++ b/docs/csi-topology.md
@@ -1,0 +1,261 @@
+# Using CSI Topology-Aware Volume Provisioning with Kubernetes
+
+## Introduction
+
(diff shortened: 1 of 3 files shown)


pure-csi/README.md
```diff
@@ -77,6 +77,9 @@ The following table lists the configurable parameters and their default values.
 | `image.tag`                 | The image tag to pull              | `5.0.8`                                   |
 | `image.pullPolicy`          | Image pull policy                  | `Always      `                            |
 | `app.debug`                 | Enable/disable debug mode for app  | `false`                                   |
+| `storagetopology.enable`    | Enable/disable csi topology feature  | `false`                                 |
```
sdodsley: This addition has broken the table

dsupure: Ah, I miss this comment. The page did not refresh until I merged. I've created another PR to fix it. #198

PR #199, opened 2020-03-31
Update docs for featuregate version changes

sdodsley:


2vcps:
Should we add the k8s docs URL to reference more info on featuregates and when they became default?
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
It is not easy to read but if you know what to look for you can see VolumeSnapshotDataSource is default and beta starting in 1.17 (for example).

sdodsley:
I could add the link with a warning it's hard to read - TMI...
What do other people think?

2vcps:
@sdodsley I see it as a catch all for the next time someone installs 1.21 on day 0 and asks why blah blah blah feature isn't working or there is an error on the enabling of a feature.

PR #202, opened 2020-04-02
Added Read-Write-Many docs

Pure-AdamuKaapan:
Adds a documentation page for RWX volumes. Also starts the convention of having an examples directory in the docs page, for easier usage of demo files.

### DIFF
--- a/docs/csi-read-write-many.md
+++ b/docs/csi-read-write-many.md
@@ -1,0 +1,61 @@
+
+# Using Read-Write-Many (RWX) volumes with Kubernetes
+
+## Introduction
+
+The Pure Service Orchestrator Kubernetes CSI driver includes support for Read-Write-Many (RWX) block volumes on FlashArray
+starting with version 5.1.0. This feature allows Kubernetes end-users to create persistent block volumes that may be mounted into
+multiple pods simultaneously. Persistent volume claims created this way can be mounted exactly the same as a normal pod, only
+requiring that `accessModes` contains `ReadWriteMany`.
+
+## Restrictions
+Read-Write-Many cannot be used with all combinations of storage classes. Specifically, we prohibit mounting a block volume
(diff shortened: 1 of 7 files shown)


docs/csi-read-write-many.md
```diff
@@ -0,0 +1,61 @@
+
+# Using Read-Write-Many (RWX) volumes with Kubernetes
+
+## Introduction
+
+The Pure Service Orchestrator Kubernetes CSI driver includes support for Read-Write-Many (RWX) volumes starting with version 5.1.0.
```
sdodsley, line 6: Makes it sound like we haven't before, even though we have with FlashBlade so maybe say `for FlashArray` somewhere

docs/csi-read-write-many.md
```diff
@@ -0,0 +1,61 @@
...
+|--------------|------------|-------------|--------|
+| Block        | Block      | RWO         | Yes    |
+| Block        | Block      | RWX         | Yes    |
+| Block        | File       | RWO         | Yes    |
+| Block        | File       | RWX         | **No** |
+| File         | File       | RWO         | Yes    |
+| File         | File       | RWX         | Yes    |
+
+## Examples
+To use these examples, install the Pure CSI plugin (either through the operator or helm).
```
sdodsley, line 25: let's leave operators out of this... We are probably going to get rid of that deployment method now OpenShift are allowing Helm again

docs/csi-read-write-many.md
Pure-AdamuKaapan: Oh thanks, I forgot that we supported it on FB before, I'll fix that.

docs/csi-read-write-many.md
Pure-AdamuKaapan: Perfect, that works for me.

PR #203, opened 2020-04-02
Fix the FlashBlade mount pending problem

dsupure:
** Summary of changes: **
* Fix the FlashBlade mount pending problem when we set `pure` as the default StorageClass and set `pureBackend = file`
* Also picked up #159 PR's change.

This branch will be tested using our release pipeline.

### DIFF
--- a/pure-csi/templates/storageclass.yaml
+++ b/pure-csi/templates/storageclass.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.storageclass.createBuiltIn }}
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
@@ -61,4 +62,5 @@ mountOptions:
 {{- else if kindIs "string" .Values.flasharray.defaultMountOpt}}
 mountOptions:
   - {{ .Values.flasharray.defaultMountOpt }}
+{{- end }}
 {{- end }}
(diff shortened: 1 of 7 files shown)


dsupure:
> Doe this default issue not affect the flex driver as well? Shouldn't the changes in the storageclass YAML for CSI be mirrored in the Flex version?

`csi.storage.k8s.io/fstype`, `createoptions` and `mountOptions`  parameters only support in the CSI version. So we don't have to mirror it to the flex version.

However, for consistency, let me uncomment `pureBackend` in the flex values.yaml  
```
storageclass:
  # create the built-in StorageClasses 'pure', 'pure-file' and 'pure-block'?
  createBuiltIn: true
  isPureDefault: false
  # set the type of backend you want for the 'pure' storageclass
  pureBackend: file
```

pure-csi/templates/storageclass.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.storageclass.createBuiltIn }}
```
caryli-ps, line 1: [Suggestion]
What's the default value for `createBuiltIn` if this field is missing from user's `values.yaml`? I think it will be false by default.

If that's the case, what do you think about changing the comparison to `createBuiltIn != false` for better backward compatibility? We would love user to refresh their `values.yaml` all the time, but in reality it's not gonna happen - and user will be stuck without the built in storage classes which they didn't intend to remove.

dsupure: Helm chart uses the value in `pure-csi/values.yaml` as the default value if users don't specify the `createBuiltIn` value in their own `values.ymal` file. The default value should be `true` and specified here already:
https://github.com/purestorage/helm-charts/pull/203/files#diff-0e144d767d619e203c878598a3f20b5aR47

The changes are from #159 and I just want to verify it in our release pipeline and close this PR. You can refer to the discussion of that PR.

PR #204, opened 2020-04-03
Add examples yaml files

dsupure:
**Summary of changes:**
* Add example yaml files for customers to quickly apply.


docs/csi-snapshot-clones.md
```diff
@@ -136,35 +136,45 @@ spec:
     name: pure-claim
     kind: PersistentVolumeClaim
 ```
-
+To give a try:
```
sdodsley, line 139: To give it a try:

docs/csi-snapshot-clones.md
```diff
@@ -136,35 +136,45 @@ spec:
...
       storage: 10Gi
-  storageClassName: pure-block
+  storageClassName: pure
   dataSource:
     kind: VolumeSnapshot
     name: volumesnapshot-1
     apiGroup: snapshot.storage.k8s.io
 ```
-
+To give a try:
```
sdodsley, line 173: same

docs/csi-snapshot-clones.md
```diff
@@ -136,35 +136,45 @@ spec:
...
-  name: pvc-restore-from-volumesnapshot-1
+  name: pvc-restore-from-volumesnapshot
 spec:
   accessModes:
     - ReadWriteOnce
   resources:
     requests:
       storage: 10Gi
-  storageClassName: pure-block
+  storageClassName: pure
```
caryli-ps, line 167: [Suggestion]
Should the storageClassName stay as `pure-block`?

I'm worrying that user would have FB setup only, and their `values.yaml` may be updated to use `file` as default backend. Then this example wouldn't work at the moment because we don't have snapshot support for FlashBlade yet.

Same comment for a few similar changes below.

sdodsley: Let's just get rid of the `pure` storage class....
As we move to the new repo for CSI that would be a good time to do that.

dsupure:
> Thanks for the document updates!
> 
> Maybe could we chat offline about the folder structure we want for the feature introduction MD file and also the example YAML files? Feels to me example YAML files should be grouped by feature (snapshot/clone/topology), rather than grouped by type (pvc/pod/snapshot/statefulset).
> 
> What do you think?

The reason I choose to group by feature is that we can reuse those yaml files, but I am fine with either way.

docs/csi-snapshot-clones.md
dsupure: @caryli-ps , good point. I did not think about the FB case. We should keep `pure-block`.  And yes, we can get rid of `pure` in the new repo.

sdodsley: This makes me 😄 .LGTM

caryli-ps:
> > Thanks for the document updates!
> > Maybe could we chat offline about the folder structure we want for the feature introduction MD file and also the example YAML files? Feels to me example YAML files should be grouped by feature (snapshot/clone/topology), rather than grouped by type (pvc/pod/snapshot/statefulset).
> > What do you think?
> 
> The reason I choose to group by feature is that we can reuse those yaml files, but I am fine with either way.

Definitely a valid point. The YAMLs are not very re-usable, unless we either give unique names to different examples, or update documents to `kubectl apply` in different namespaces.

dsupure:
> > > Thanks for the document updates!
> > > Maybe could we chat offline about the folder structure we want for the feature introduction MD file and also the example YAML files? Feels to me example YAML files should be grouped by feature (snapshot/clone/topology), rather than grouped by type (pvc/pod/snapshot/statefulset).
> > > What do you think?
> > 
> > 
> > The reason I choose to group by feature is that we can reuse those yaml files, but I am fine with either way.
> 
> Definitely a valid point. The YAMLs are not very re-usable, unless we either give unique names to different examples, or update documents to `kubectl apply` in different namespaces.

Cool, I already changed it to feature-based examples.

Issue #178, opened 2020-02-24
node-driver-registrar preStop hook: stat /bin/sh: no such file or directory


javefang:
Issue:

pure-csi (daemonset) pod failed to stop due to preStop hook exec failure. (stat /bin/sh: no such file or directory).

Cause:

The preStop hook is trying to run on image "quay.io/k8scsi/csi-node-driver-registrar:v1.1.0"
https://github.com/purestorage/helm-charts/blob/5.0.7/pure-csi/templates/node.yaml#L84-L90
which stopped including /bin/sh since version v1.1.0.

How to reproduce:

1. Install the pure-csi driver 5.0.6
2. Wait for pure-csi daemonset pod to start
3. Delete one of the pure-csi pod
4. "k describe pod pure-csi-<id>"

OR

```
dockr run --rm -ti --entrypoint /bin/sh quay.io/k8scsi/csi-node-driver-registrar:v1.1.0 
```

Environment:

OS: Centos 7
Kubernetes: v1.14.5
Pure-csi: 5.0.6 (also a problem for 5.0.7)

sdodsley:
@javefang after trawling through some repos and readme's it looks like the `preStop` hook is actually redundant. 
All we should need to do is remove the hook. Do you agree with this assessment?

javefang:
Hi @sdodsley , thanks for the swift reply! Looking at https://github.com/kubernetes-csi/node-driver-registrar/issues/21, it feels like the maintainer knows about this and is thinking of replacing the preStop hook with a SIGTERM handler. But the issue is still open.

It is probably ok to remove the preStop hook here as it is: 1) not serving the purpose 2) causing pod stop failure. Do you know if the driver can restart fine if the two paths/files that the hook was trying to delete already exist?

sdodsley:
I'll check this out and get back to you.

sdodsley:
Reolved

PR #206, opened 2020-04-07
Upgrade csi-node-driver-registrar to v1.3.0 and remove preStop hook

dsupure:
Fix the issue https://github.com/purestorage/helm-charts/issues/178

### DIFF
--- a/pure-csi/templates/node.yaml
+++ b/pure-csi/templates/node.yaml
@@ -81,13 +81,9 @@ spec:
       containers:
         - name: node-driver-registrar
           {{- with .Values.csi.nodeDriverRegistrar.image }}
-          image: {{ .name | default "quay.io/k8scsi/csi-node-driver-registrar" }}:v1.1.0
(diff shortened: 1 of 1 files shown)


pure-csi/templates/node.yaml
```diff
@@ -81,13 +81,9 @@ spec:
       containers:
         - name: node-driver-registrar
           {{- with .Values.csi.nodeDriverRegistrar.image }}
-          image: {{ .name | default "quay.io/k8scsi/csi-node-driver-registrar" }}:v1.1.0
+          image: {{ .name | default "quay.io/k8scsi/csi-node-driver-registrar" }}:v1.3.0
           imagePullPolicy: {{ .pullPolicy }}
           {{- end }}
-          lifecycle:
```
pure-jliao, line 87: should those also be removed in DB branch? I didn't remove those.

dsupure: Yeah, it would be nice to remove it as well.

PR #213, opened 2020-04-14
Fix incorrect OpenShift support version

sdodsley:
The CSI driver doesn't work in OpenShift 3.11 and we don't have a CI for OSE 4.3 so remove this (temporarily)

### DIFF
--- a/pure-csi/README.md
+++ b/pure-csi/README.md
@@ -12,7 +12,6 @@ This helm chart installs the CSI plugin on a Kubernetes cluster.
 - #### Environments Supported*:
   - Kubernetes 1.13+
   - Minimum Helm version required is 3.1.0.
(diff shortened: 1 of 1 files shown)


dsupure:
Do we have anyone or any customers tested 4.3?

sdodsley:
This is my issue - we apparently do, but I don't know what tweaks they had to make. I'm trying to build OpenShift 4.3 myself but with little success.
We definitely shouldn't have 3.11 listed as that doesn't support CSI

sdodsley:
OK - So until we have a CI for OSE 4.3 I've removed the OpenShift support statement for non-operator CSI

dsupure:
Sounds good! We should add OSE 4.3 test as soon as we can.

PR #218, opened 2020-04-21
Add volume import readme and examples

haibinxie:


docs/csi-volume-import.md
```diff
@@ -0,0 +1,74 @@
...
+  csi:
+    driver: pure-csi
+    # TODO: change to the volume name in backend.
+    # Volume with any name that exists in backend can be imported, and will not be renamed.
+    # Alternatively for stateful CSI, this can be volume id in persistent store.
+    volumeHandle: ttt-pvc-a90d7d5f-da6c-44db-a306-a4cc122f9dd3
+  # TODO: configure your desired reclaim policy,
+  # Use Retain if you don't want your volume to get deleted when the PV is deleted.
+  persistentVolumeReclaimPolicy: Delete
+  storageClassName: pure-file
```
sdodsley, line 45: Shouldn't this be `pure-block`? We can't do soft import for FB until the db branch is live.

docs/csi-volume-import.md
```diff
@@ -0,0 +1,74 @@
...
+  # Note: These two fields are not required for pre-bound PV.
+  # storageClassName: pure-block
+  # volumeMode: Filesystem
+
+  # TODO: Change to the name of the imported PV.
+  volumeName: pv-import
+```
+
+3. Use the persistent volume claim.
+
```
sdodsley, line 73: example 3 is missing

docs/csi-volume-import.md
```diff
@@ -0,0 +1,74 @@
...
+If users want to import a volume that is created outside of Kubernetes, persistent volume and persistent volume claim objects can be manually created following below steps:
+
+1. Create and deploy a persistent volume object with volumeHandle configured to name of the volume in backend, claimRef configured to the name of the persistent volume claim at your choice.
+
+    Here is an example, more examples can be found at: examples/apps/volumeimport/...
+```yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  annotations:
```
sdodsley, line 20: is this annotation necessary?

docs/csi-volume-import.md
```diff
@@ -0,0 +1,74 @@
...
+    # TODO: change to the PVC you want to bind this PV.
+    # If you don't pre-bind PVC here, the PV might be automatically bound to a PVC by scheduler.
+    name: pvc-import
+    # Namespace of the PVC
+    namespace: default
+  csi:
+    driver: pure-csi
+    # TODO: change to the volume name in backend.
+    # Volume with any name that exists in backend can be imported, and will not be renamed.
+    # Alternatively for stateful CSI, this can be volume id in persistent store.
```
sdodsley, line 40: we shouldn't have this comment in the current branch only in the db branch.
If someone sees this they may well get confused

docs/examples/volumeimport/pv-import-block.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  annotations:
```
sdodsley, line 4: Needed?

docs/examples/volumeimport/pv-import-file.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
```
sdodsley, line 1: This example should be in this repo as it requires the stateful version

docs/csi-volume-import.md
haibinxie: soft import with master branch also works.

docs/csi-volume-import.md
haibinxie: I did not intend to give example originally since it's a regular PVC, but I will add one here to make it easier to read.

docs/csi-volume-import.md
haibinxie: yes, it tells Kubernetes to use our CSI plugin.

docs/csi-volume-import.md
haibinxie: Thanks for pointing out, will correct shortly

docs/csi-volume-import.md
sdodsley: How without the db lookup as re can't rename the FB share- did I miss something?

sdodsley: Super excited if this is so

docs/examples/volumeimport/pv-import-file.yaml
haibinxie: The stateless version works for FA. is there any reason why it would not work

sdodsley: We need to change the volume/share name if there is no DB to do a lookup - you can't rename shares on a FB

haibinxie: We don't rename in both stateful and stateless versions, if the volume does not exist in backend it would fail

docs/csi-volume-import.md
sdodsley: if there are multiple backends and they share a share name how does PSO differentiate? If the backends are FBs then there is no option to rename so we have to be able to identify which backend

haibinxie: the original intention was to let it fail if we find multiple volumes with same name, I will add code to check this (in master branch)

docs/csi-volume-import.md
```diff
@@ -0,0 +1,93 @@
+## Import volumes into Kubernetes
+
+#### Use Cases
+- Reinstall Kubernetes or migrate volume from one Kubernetes cluster to another
+- Disaster recovery from backup volume
+- Containerize legacy application
+
+#### Import Guidance
```
sdodsley, line 8: I think we should add something in here about `retain` and `delete`. It's pretty important and critical.

haibinxie: sounds good, added

docs/csi-volume-import.md
```diff
@@ -0,0 +1,100 @@
...
+```
+>Both Delete and Retain reclaim policies are supported on imported volumes, reclaim policy is configured in persistent volume object. The backend volume will be deleted automatically in below scenarios:
+>1. Users delete a persistent volume claim and the corresponding persistent volume has Delete as reclaim policy. In this case persistent volume object and backend volume will be deleted automatically.
+>2. Users delete a persistent volume when it is not bound to any PVC. In this case backend volume will be deleted automatically.
+
+If users want to import a volume that is created outside of Kubernetes, persistent volume and persistent volume claim objects can be manually created following below steps:
+
+1. Create and deploy a persistent volume object with volumeHandle configured to name of the volume in backend, claimRef configured to the name of the persistent volume claim at your choice.
+
+    Here is an example, more examples can be found at: examples/apps/volumeimport/...
```
dsupure, line 22: For `examples/apps/volumeimport/...` can we use a hyperlink here?

docs/csi-volume-import.md
```diff
@@ -0,0 +1,100 @@
+# Import volumes into Kubernetes
+
+## Use Cases
```
dsupure, line 3: It would be nice to have an **Introduction** and **Dependencies** sections like this: 
https://github.com/purestorage/helm-charts/blob/master/docs/csi-topology.md
I think you need to mention that this feature starts to support from v5.2.0+.

docs/csi-volume-import.md
sdodsley: Agreed

sdodsley:
We should not be adding this to master until the 5.1.1 release has merged and the changes must reference 5.2.0 as the minimum version as @dsupure  mentioned.

haibinxie:
> Wait until 5.1.1 merges

Sounds good

docs/csi-volume-import.md
haibinxie: fixed

docs/csi-volume-import.md
haibinxie: I added dependencies, the name and use cases introduce the feature pretty well, there might be little help to add an introduction section.

dsupure: Yeah, no worries. I think it is good as long as we have **Dependencies** section.

docs/examples/volumeimport/pv-import-raw.yaml
```diff
@@ -0,0 +1,29 @@
...
+    namespace: default
+  csi:
+    driver: pure-csi
+    # TODO: change to the volume name in backend.
+    # Volume with any name that exists in backend can be imported, and will not be renamed.
+    volumeHandle: ns04132-pvc-540d6142-2e86-45ba-939d-c0be5d8fd335
+  # TODO: configure your desired reclaim policy,
+  # Use Retain if you don't want your volume to get deleted when the PV is deleted.
+  persistentVolumeReclaimPolicy: Delete
+  storageClassName: pure-block
```
sdodsley, line 28: shouldn't this be a different sc given you are importing a raw volume?

docs/csi-volume-import.md
```diff
@@ -0,0 +1,104 @@
...
+- Reinstall Kubernetes or migrate volume from one Kubernetes cluster to another
+- Disaster recovery from backup volume
+- Containerize legacy application
+
+## Dependencies
+* Kubernetes v1.15+ (Lower version might work however not tested)
+* Pure CSI driver 5.2.0+
+
```
sdodsley, line 11: Might be worth stating clearly that import is supported for FA (raw and file) and FB (file)

docs/csi-volume-import.md
```diff
@@ -0,0 +1,104 @@
...
+## Import Guidance
+The beauty of volume import feature is there are no changes or additional annotations to persistent volume objects.  Take the scenario of migrating volumes from one Kubernetes cluster to another cluster for example, users can export persistent volume objects (and persistent volume claim objects) from old Kubernetes cluster and deploy to new Kubernetes cluster, no additional changes required.
```
sdodsley, line 13: Grammar points:
* The beauty of the volume import is that there are no...
* `for example` not required in the second sentence and use a colon after `cluster` and before `users`
* from the old Kubernetes cluster and deploy to a new Kubernetes cluster, with no additional...

docs/csi-volume-import.md
```diff
@@ -0,0 +1,104 @@
...
+
+```diff
+-Caution: important notes about reclaim policy
+```
+>Both Delete and Retain reclaim policies are supported on imported volumes, reclaim policy is configured in persistent volume object. The backend volume will be deleted automatically in below scenarios:
```
sdodsley, line 18: volumes, where the reclaim policy...
change `below` to `following`

docs/csi-volume-import.md
```diff
@@ -0,0 +1,104 @@
...
+>1. Users delete a persistent volume claim and the corresponding persistent volume has Delete as reclaim policy. In this case persistent volume object and backend volume will be deleted automatically.
```
sdodsley, line 19: highlight Delete as `Delete`
In this case, both the persistent...

docs/csi-volume-import.md
```diff
@@ -0,0 +1,104 @@
...
+>2. Users delete a persistent volume when it is not bound to any PVC. In this case backend volume will be deleted automatically.
```
sdodsley, line 20: In this case the backend volume...

docs/csi-volume-import.md
```diff
@@ -0,0 +1,104 @@
...
+
+If users want to import a volume that is created outside of Kubernetes, persistent volume and persistent volume claim objects can be manually created following below steps:
```
sdodsley, line 22: was created, not is created
using the following steps:

docs/csi-volume-import.md
```diff
@@ -0,0 +1,104 @@
...
+
+1. Create and deploy a persistent volume object with volumeHandle configured to name of the volume in backend, claimRef configured to the name of the persistent volume claim at your choice.
```
sdodsley, line 24: highlight volumeHandle as `volumeHandle`
configured to be the...
change claimRef to be   .... and `claimRef` ...

docs/csi-volume-import.md
```diff
@@ -0,0 +1,104 @@
...
+
+    Here is an example, more examples can be found at: [examples/volumeimport](./examples/volumeimport)
```
sdodsley, line 26: Here is an example, but more examples....

docs/examples/volumeimport/pv-import-raw.yaml
haibinxie: there are only two sc types, file and block, we kind of have 3 storage classes in our deployments by default: pure, pure-file and pure-block, and in fact pure is same to pure-block. raw block and file system on block both uses same pure-block storage class.

sdodsley: Got it - I got a little confused on the way we create raw devices... we do use just the `pure-block` but have a different `volumeMode`.

docs/csi-volume-import.md
haibinxie: all resolved.

docs/csi-volume-import.md
```diff
@@ -0,0 +1,105 @@
...
+```diff
+-Caution: important notes about reclaim policy
+```
+>Both Delete and Retain reclaim policies are supported on imported volumes, where the reclaim policy is configured in persistent volume object. The backend volume will be deleted automatically in following scenarios:
+>1. Users delete a persistent volume claim and the corresponding persistent volume has `Delete` as reclaim policy. In this case, both the persistent volume object and backend volume will be deleted automatically.
+>2. Users delete a persistent volume when it is not bound to any PVC. In this case the backend volume will be deleted automatically.
+
+If users want to import a volume that was created outside of Kubernetes, persistent volume and persistent volume claim objects can be manually created using the following steps:
+
+1. Create and deploy a persistent volume object with `volumeHandle` configured to be the name of the volume in backend, claimRef configured to the name of the persistent volume claim at your choice.
```
sdodsley, line 25: climRef to `claimRef`

haibinxie:
@sdodsley I updated the notes about reclaim policy, actually if users delete persistent volume object, the backend volume will not be deleted regardless what reclaim policy is used, i think Kubernetes was designed this way to give people a chance to not delete volumes with Delete reclaim policy if they change mind.

sdodsley:
That makes no sense. We use `delete` by default in our storage classes and when we delete the PV or PVC the backend volume always gets deleted.

haibinxie:
> That makes no sense. We use `delete` by default in our storage classes and when we delete the PV or PVC the backend volume always gets deleted.

I dont think that's the case, i have tested it and indeed our test function has been using this way to create a volume in backend in many places. you can try real quick on your environment.

docs/csi-volume-import.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Import Guidance
+The beauty of volume import feature is that there are no changes or additional annotations to persistent volume objects.  Take the scenario of migrating volumes from one Kubernetes cluster to another cluster: users can export persistent volume objects (and persistent volume claim objects) from the old cluster and deploy to a new Kubernetes cluster, with no additional changes.
+
+```diff
+-Caution: important notes about reclaim policy
+```
+>Both Delete and Retain reclaim policies are supported on imported volumes, where the reclaim policy is configured in persistent volume object. If users delete a persistent volume claim and the corresponding persistent volume has `Delete` as reclaim policy, both the persistent volume object and backend volume will be deleted automatically.
+>
+>If users delete a persistent volume object (via kubectl command or Kubernetes API), the backend volume will *NOT* be deleted regardless of whether the reclaim policy is Delete or Retain. The behavior is same to dynamically provisioned volumes.
```
sdodsley, line 21: Highlight this piece, so the result depending on the order things are deleted, PVC then PV, PV then PVC, will dramatically change the outcome.

PR #222, opened 2020-04-25
Update volume expansion with better kubectl commands

sdodsley:
Better to use `kubectl patch` that `kubectl edit`.
Also:
* correct the markdown formatting
* make the comment about pod restart more prominent
* remove redundant YAML text

### DIFF
--- a/docs/csi-volume-expansion.md
+++ b/docs/csi-volume-expansion.md
@@ -16,7 +16,10 @@ The following prerequisites must be true to ensure the function can work properl
 * Only allow volume size expansion, shrinking a volume is not allowed.
 * The `StorageClass` that provisions the PVC must support resize. `allowVolumeExpansion` flag is set to true by default in all PSO `StorageClass` since version 5.2.0.
 * The PVC `accessMode` must be `ReadWriteOnce` or `ReadWriteMany`.
-* For PSO `pure-block` `StorageClass`, the filesystem/block on node is only resized upon a Pod (re-)start. 
+
+### Note
+
+Any PVC creted using a StorageClass where the `Parameters:   backend=block` is true will only be resized upon a (re)start of the pod bound to the PVC. 
 
 ## Dependencies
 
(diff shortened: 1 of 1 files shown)


docs/csi-volume-expansion.md
```diff
@@ -167,5 +175,17 @@ kubectl apply -f https://raw.githubusercontent.com/purestorage/helm-charts/maste
...
-kubectl edit pvc pure-claim-file
+kubectl patch pvc pure-claim-file -p='{"spec": {"resources": {"requests": {"storage": "20Gi"}}}}'
+```
+
+Check that both the PV and PVC are immeadiately expanded. No pod restart is required.
+
+```bash
+# kubectl get pvc pure-claim-file
+NAME               STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
+pure-claim-file    Bound     pvc-2ba56b33-3412-2965-f4e4-983de21ba772   10Gi       RWO            pure-file      25s
```
pure-jliao, line 186: Here the size should be 20?

docs/csi-volume-expansion.md
```diff
@@ -167,5 +175,17 @@ kubectl apply -f https://raw.githubusercontent.com/purestorage/helm-charts/maste
...
+# kubectl get pv
+NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS    CLAIM                      STORAGECLASS   REASON    AGE
+pvc-2ba56b33-3412-2965-f4e4-983de21ba772   20Gi       RWX            Delete           Bound     default/pure-claim-file    pure-file                27s
```
pure-jliao, line 189: Access mode should be RWO?

docs/csi-volume-expansion.md
sdodsley: Yup - typo

docs/csi-volume-expansion.md
sdodsley: `pure-file` is really only RWX. Are you saying that RWX cannot be expanded? Even RWX on raw block?

pure-jliao: Oh didn’t know. I mean the pvc is RWO but the PV is RWX?

sdodsley: Made it RWO, but let's confirm about RWX expansion support on `pure-file` and raw block in `pure-block`.

pure-jliao: Confirmed it should work for both.

docs/csi-volume-expansion.md
pure-jliao: here the issue is not resolved.

caryli-ps: Line 186 CAPACITY needs to be updated to show `20Gi` as well.

PR #224, opened 2020-05-06
Updated volume import docs with backend parameters

Pure-AdamuKaapan:
For very specific cases involving ReadWriteMany (RWX) access modes on imported volumes, certain parameters are missing from the CSI requests. This fixes that issue by manually providing the important parameter (`backend`).


haibinxie:
LGTM

PR #230, opened 2020-05-27
[WIP] Test for v1beta1 snapshot API

sdodsley:
Only for testing against internal CI system


pure-csi/templates/rbac.yaml
```diff
@@ -51,6 +51,9 @@ rules:
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents"]
     verbs: ["create", "get", "list", "watch", "update", "delete"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
```
pure-jliao, line 54: might need this change in operator install.sh as well.

sdodsley: Good catch. Done

pure-csi/templates/provisioner.yaml
```diff
@@ -86,12 +86,11 @@ spec:
 {{- if not (.Capabilities.KubeVersion.GitVersion | regexMatch "gke") }}
         - name: csi-snapshotter
           {{- with .Values.csi.snapshotter.image }}
-          image: {{ .name | default "quay.io/k8scsi/csi-snapshotter" }}:v1.2.2
+          image: {{ .name | default "quay.io/k8scsi/csi-snapshotter" }}:v2.1.1
```
pure-jliao, line 89: as described in the release note https://github.com/kubernetes-csi/external-snapshotter/releases
from `v2.0.0` requries k8s 1.17 and above, I think we should keep `v1.2.2` for older k8s versions.

sdodsley: This is just to test if things work in 1.17. Will add version checks if they do

sdodsley:
Closing this in favour of a more specific OpenShift 4.4 PR

PR #242, opened 2020-07-15
[NO-ISSUE] Anthos 1.4 support and k8s version check changes

pure-jliao:
* smoke passed, should be safe: http://newstack-jenkins/blue/organizations/jenkins/k8s/detail/master/854/pipeline/219/

### DIFF
--- a/pure-csi/README.md
+++ b/pure-csi/README.md
@@ -12,7 +12,7 @@ This helm chart installs the CSI plugin on a Kubernetes cluster.
 - #### Environments Supported*:
   - Kubernetes 1.13+
   - Minimum Helm version required is 3.1.0.
-  - Google Anthos 1.2.x, 1.3.x
+  - Google Anthos 1.2.x, 1.3.x, 1.4.x
   - Docker Kuberenetes Service (DKS) - based on Docker EE 3.0 with Kubernetes 1.14.3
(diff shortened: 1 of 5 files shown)


david-enli:
I'm curious
what does this mean? 
`# Google Anthos (which is built on GKE) prohibits alpha snapshotters`
1.2.2 is a alpha snapshotter tho

pure-jliao:
> I'm curious
> what does this mean?
> `# Google Anthos (which is built on GKE) prohibits alpha snapshotters`
> 1.2.2 is a alpha snapshotter tho

My understand is Anthos prohibits alpha features, so we don't enable feature-gates manually on the cluster
from release note https://github.com/kubernetes-csi/external-snapshotter/releases/tag/v1.2.2
>Volume snapshotting is an Alpha feature in Kubernetes v1.15.
it doesn't say anything about the snapshotter is alpha, since gke is upgraded to 1.16, I'm checking if we could run our functional test on it, the clone feature is added and requires the snapshotter if I'm correct.

pure-jliao:
@david-enli is correct, now it's clear. our snapshot class is using v1alpha1, and that is truly prohibited.
`alpha-features VolumeSnapshot alpha CRD should not be deployed to the cluster`

removing the snapshotter and keep the version check fixes only.

PR #250, opened 2020-08-20
[NO-ISSUE] Added warnings linking to new PSO 6 repo

Pure-AdamuKaapan:


Pure-AdamuKaapan:
(note: both operators already had notes about when to use them, and the flex chart already has a note saying not to use it)

PR #253, opened 2020-10-15
[NO ISSUE] Change pod and DS name for GKE

sdodsley:
Change the name of the pod and DS to something less confusing...

### DIFF
--- a/pure-csi/templates/node-configure.yaml
+++ b/pure-csi/templates/node-configure.yaml
@@ -3,7 +3,7 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: pso-gke-config
+  name: pso-node-config
   namespace: {{ .Release.Namespace }}
 spec:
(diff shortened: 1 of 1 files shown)


pure-csi/templates/node-configure.yaml
```diff
@@ -3,20 +3,20 @@
...
 kind: DaemonSet
 metadata:
-  name: pso-daemonset
+  name: pso-gke-config
   namespace: {{ .Release.Namespace }}
 spec:
   selector:
     matchLabels:
-      name: pso-daemonset
+      name: pso-node-config
```
pure-jliao, line 11: don't we make the name consistent here? sorry not super familiar how this works..

sdodsley: Yep - I missed the entry on line 6. Fixed now.

pure-jliao (approved):

PR #257, opened 2021-04-05
Add max k8s version

sdodsley:
A customer requested the max k8s version supported to be added.
Given PSO 5 only supports the Alpha snapshot API we have to limit the max to when this changed to beta, ie 1.17
Also change maximum versions for the old Flex driver plugins as well


Pure-AdamuKaapan (approved):
LGTM, thanks for adding this, should help clear some things up

PR #259, opened 2021-04-22
Bolt door after horse has bolted...

sdodsley:
Specify no use of hyphens in the `namespace.pure` parameter


dsupure (approved):
Like your title.

Pure-AdamuKaapan (approved):
LGTM

